PR #42, opened 2015-03-23
Stability fixes for the message loop

am0s:
Several fixes to make the socket client message loop more resilient against exceptions/errors.

Also includes a utf-8 fix when sending json messages.

### DIFF
--- a/pychromecast/socket_client.py
+++ b/pychromecast/socket_client.py
@@ -183,6 +183,11 @@ class SocketClient(threading.Thread):
 
         return self._request_id
 
+    @property
+    def is_stopped(self):
+        """ Returns True if the connection has been stopped, False if it is running. """
+        return self.stop.is_set()
+
     def run(self):
         """ Start polling the socket. """
         self.receiver_controller.update_status()
--- a/pychromecast/socket_client.py
+++ b/pychromecast/socket_client.py
@@ -237,10 +237,16 @@ class SocketClient(threading.Thread):
                     event.set()
 
         for handler in self._handlers.values():
-            handler.tear_down()
+            try:
+                handler.tear_down()
+            except Exception:  # pylint: disable=broad-except
+                pass
 
         for channel in self._open_channels:
-            self._disconnect_channel(channel)
+            try:
+                self._disconnect_channel(channel)
+            except Exception:  # pylint: disable=broad-except
+                pass
 
         self.socket.close()
 
--- a/pychromecast/socket_client.py
+++ b/pychromecast/socket_client.py
@@ -260,7 +260,7 @@ class SocketClient(threading.Thread):
     # pylint: disable=too-many-arguments
     def send_message(self, destination_id, namespace, data,
                      inc_session_id=False, wait_for_response=False,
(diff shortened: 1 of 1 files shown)


pychromecast/socket_client.py
```diff
@@ -281,7 +301,7 @@ def send_message(self, destination_id, namespace, data,
         msg.destination_id = destination_id
         msg.payload_type = cast_channel_pb2.CastMessage.STRING
         msg.namespace = namespace
-        msg.payload_utf8 = json.dumps(data, ensure_ascii=False).encode("utf8")
+        msg.payload_utf8 = json.dumps(data)
```
balloob: This breaks compatibility with Python 3.

Although Python 3 support is currently broken, it is the plan to support it as soon as Google releases version 3 of protobuf (which includes Python 3 support).

pychromecast/socket_client.py
```diff
@@ -204,11 +213,16 @@ def run(self):
...
+                        message, data)
 
-                if not handled:
-                    self.logger.warning("Message unhandled: {}".format(
+                    if not handled:
+                        self.logger.warning("Message unhandled: {}".format(
+                            _message_to_string(message, data)))
+                except Exception, e:
+                    self.logger.exception(u"Exception {} caught while sending message to controller {}: {}".format(
+                        unicode(e), type(self._handlers[message.namespace]).__name__,
```
balloob: unicode(e) is not python 3 compatible either. Shouldn't format convert it to string automatically?

balloob:
Thanks for your contribution, this is awesome. If you could fix the Python 3 incompatibilities it is good to go for merging.

am0s:
Sure no problem, I'll have a look.

pychromecast/socket_client.py
am0s: Are you sure this will break?

Won't the protobuf library support strings as well as bytes in the payload_utf8 field?
ie. it will convert (unicode) strings itself?

After testing the json output a bit I think it might be an idea to change the dumps to:

    json.dumps(data, ensure_ascii=False)

That way unicode characters will be encoded to utf-8 and not encoded as the ASCII equivalent, this will be more space efficient.

pychromecast/socket_client.py
am0s: Yes, you'right, tested it a bit now. In Python 2.x as long as the source string is unicode it will convert things correctly.

Would it be a good idea to prefix all strings (that are not meant to by bytes) with the u literal?

pychromecast/socket_client.py
balloob: I am not sure if it expects or Strings. Let's go with your proposed solution for now. We'll have to do a full regression anyway when Protobuf 3 gets released.

pychromecast/socket_client.py
balloob: I prefer the other way around (a b prefix for bytes) as we're having less bytes then strings.

am0s:
I fixed the Python 3 compatibily issues. How does it look now?

balloob:
Perfect!

PR #44, opened 2015-05-28
Python3-support using protobuf-3.0.0a2

wind-rider:
Update protobuf to version 3
Generate new protobuf-scripts from https://chromium.googlesource.com/chromium/src.git/+/master/extensions/common/api/cast_channel
Update socket-reading code in socket_client to work with python3 (byte-arrays are not strings in python3)


wind-rider:
With this code, I'm able to run example.py again on python3. I only got one warning message:
`````
Received unknown namespace: Message urn:x-cast:com.google.cast.tp.connection from web-1 to sender-0: {'type': 'CLOSE'}
`````
However, functionally it works well.

I included the .proto-files I used to generate the protobuf-scripts. I imported them from the master branch at https://chromium.googlesource.com/chromium/src.git/+/master/extensions/common/api/cast_channel/. Maybe there is a better way to include them and keep them up-to-date?

wind-rider:
I excluded two more generated protobuf scripts from pylint. The .proto files for them were in https://chromium.googlesource.com/chromium/src.git/+/master/extensions/common/api/cast_channel/, not sure if they are needed.

balloob:
Awesome!

I'll hope to have time to look into it tonight. If this works I'll make sure to release a new version and get it working in HA again.

balloob:
Published new version to [PyPi](https://pypi.python.org/pypi/PyChromecast) :smile:

wind-rider:
Great to hear :D

wind-rider:
I'm busy making it work at least partly again in Home Assistant. I'll make a commit there later tonight.

I could use some help with connecting the broken parts again. For example lots of media status properties like album, image etc disappeared and also the play_next_track() method disappeared.

PR #45, opened 2015-05-31
Expose mute function

wind-rider:
I exposed your mute function (so that it can be called from Home Assistant), but that will require another upload to pypi..

Can you have a look at https://github.com/balloob/pychromecast/blob/45c7dff0600c87be0bb8a0d5a1618a7897f232da/pychromecast/__init__.py#L166-167 whether and how those functions should be exposed?

Since I'm not familiar with Python I'm not sure what is a good practice (in C# this "just adding functions to an object" like how it's done with set_volume is not possible).

set_volume is not in the Chromecast object signature, but I use it from https://github.com/balloob/home-assistant/blob/dev/homeassistant/components/media_player/cast.py#L165

I added set_volume_muted to the Chromecast object signature now.

### DIFF
--- a/pychromecast/__init__.py
+++ b/pychromecast/__init__.py
@@ -164,6 +164,7 @@ class Chromecast(object):
 
         # Forward these methods
         self.set_volume = self.socket_client.receiver_controller.set_volume
+        self.set_volume_muted = self.socket_client.receiver_controller.set_volume_muted
         self.play_media = self.socket_client.media_controller.play_media
         self.register_handler = self.socket_client.register_handler
 
@@ -224,6 +225,11 @@ class Chromecast(object):
         """
         volume = round(self.status.volume_level, 1)
         return self.set_volume(volume - 0.1)
+
+    def volume_mute(self, mute):
+        """ Mute the volume if mute is true
+        or unmute if the mute parameter is false. """
+        self.set_volume_muted(mute)
 
     def __del__(self):
         self.socket_client.stop.set()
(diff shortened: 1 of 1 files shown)


pychromecast/__init__.py
```diff
@@ -225,6 +226,11 @@ def volume_down(self):
         volume = round(self.status.volume_level, 1)
         return self.set_volume(volume - 0.1)
 
+    def volume_mute(self, mute):
```
balloob: This one is not needed since you've done
```python
self.set_volume_muted = self.socket_client.receiver_controller.set_volume_muted
```

balloob:
I feel like I need to re-organize PyChromecast, it's becoming a mess. For now, let's just stick to the first statement (`self.set_volume_muted = …`)

pychromecast/__init__.py
wind-rider: Ok, then I'll remove that volume_mute function. The other one is still needed I think (until pychromecast is refactored).

balloob:
Looks like PyLint found one style issue:
`pychromecast/__init__.py:167:80: E501 line too long (87 > 79 characters)`

pychromecast/__init__.py
```diff
@@ -164,6 +164,7 @@ def __init__(self, host, tries=None):
 
         # Forward these methods
         self.set_volume = self.socket_client.receiver_controller.set_volume
+        self.set_volume_muted = self.socket_client.receiver_controller.set_volume_muted
```
balloob:
```python
self.set_volume_muted = \
    self.socket_client.receiver_controller.set_volume_muted
```

wind-rider:
This should fix it. Now I'm off to sleep :) See you later!

balloob:
Thanks and talk to you later !

Issue #50, opened 2015-07-04
is_active_input problems


rmkraus:
I have been trying to debug some issues with Home Assistant and I traced it back down to the is_active_input parameter in the settings. I am not entirely sure what this value is supposed to represent, but I am assuming that it is intended to report if the Chromecast is on the TV's selected input.

For me, it appears that the value reported from the Chromecast is changing at random for systems that I believe to support CEC. Sometimes it works, other times it doesn't. For systems that I know do not support CEC, it appears to not be provided.

This value eventually propagates down to the is_idle property. That value is then used inside of Home Assistant for the Cast's status. When this value is not properly functioning, and an unsupported application (Netflix) is playing, HA will report Off instead of Unknown.

I don't know what anyone else's experience is with this parameter, but, assuming I have the proper understanding of its purpose, it seems buggy for me. I would contend that maybe this parameter should not be considered when calculating is_idle since maybe this parameter should only answer the question of is the Chromecast itself idle. I'm not sure it would be best to have this parameter indicate the status of other devices it is connected to.

Then again, I may not be correctly interpreting the purpose of this value.

balloob:
`is_active_input` is a value that is returned by the Chromecast. It is my understanding that it is `true` when the TV is turned on and tuned into the Chromecast channel.

I have had no problems with it in the past.

The current implementation of `is_idle` is this:
```python
    @property
    def is_idle(self):
        """ Returns if there is currently an app running. """
        return (self.status is None or not self.status.is_active_input or
                self.app_id in (None, IDLE_APP_ID))
```

`is_active_input` is one of the 3 checks to determine if the Chromecast is idle. We check:

1. If we don't have an active status
2. If we are not the active input
3. If there is no app running or the backdrop app is running

I think that we need to do some more research on the reliability of `is_active_input`. If it really is unstable we can remove the `is_idle` dependency on it.

rmkraus:
In my local copy, I've remove the "not self.status.is_active_input" check.
It gives me the expected behavior. On XDA Developers, there appears to be a
handful of people having CEC problems
<https://www.google.com/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=chromecast+cec+problems+site:forum.xda-developers.com>.
It is hard to diagnose this issue specifically as I'm not sure anyone has
really looked into this value all that closely. It appears that it is
likely a firmware issue on my devices.

At either rate, this check for is_active_input is probably not necessary
[...]

PR #51, opened 2015-07-10
Added Ignore CEC Capability

rmkraus:
Added the ability to ignore CEC information from the Chromecast if it is returning incorrect data. This is to address issue #50.

Also, [happy birthday Paulus!](https://www.youtube.com/watch?v=KKfNYmNdiF0)


balloob:
Thanks!

Issue #47, opened 2015-06-01
Can't import pychromecast 0.6.4 on Python 2.7.9


bowans:
In my test, Python 3.4.3 works fine.
But fail to import on Python 2.7.9.

Install pychromecast through `pip install pychromecast`
`pip list` -> `pychromecast (0.6.4)`

```
Python 2.7.9 (default, Dec 10 2014, 12:24:55) [MSC v.1500 32 bit (Intel)] on win
32
Type "help", "copyright", "credits" or "license" for more information.
>>> import pychromecast
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Python27\lib\site-packages\pychromecast\__init__.py", line 11, in <mo
dule>
    from . import socket_client
  File "C:\Python27\lib\site-packages\pychromecast\socket_client.py", line 20, i
n <module>
    from . import cast_channel_pb2
  File "C:\Python27\lib\site-packages\pychromecast\cast_channel_pb2.py", line 23
, in <module>
    serialized_pb=_b('\n\x12\x63\x61st_channel.proto\x12 extensions.core_api.cas
t_channel\"\xed\x02\n\x0b\x43\x61stMessage\x12W\n\x10protocol_version\x18\x01 \x
02(\x0e\x32=.extensions.core_api.cast_channel.CastMessage.ProtocolVersion\x12\x1
1\n\tsource_id\x18\x02 \x02(\t\x12\x16\n\x0e\x64\x65stination_id\x18\x03 \x02(\t
\x12\x11\n\tnamespace\x18\x04 \x02(\t\x12O\n\x0cpayload_type\x18\x05 \x02(\x0e\x
32\x39.extensions.core_api.cast_channel.CastMessage.PayloadType\x12\x14\n\x0cpay
load_utf8\x18\x06 \x01(\t\x12\x16\n\x0epayload_binary\x18\x07 \x01(\x0c\"!\n\x0f
ProtocolVersion\x12\x0e\n\nCASTV2_1_0\x10\x00\"%\n\x0bPayloadType\x12\n\n\x06STR
ING\x10\x00\x12\n\n\x06\x42INARY\x10\x01\"s\n\rAuthChallenge\x12\x62\n\x13signat
ure_algorithm\x18\x01 \x01(\x0e\x32\x34.extensions.core_api.cast_channel.Signatu
[...]

PR #52, opened 2015-07-16
Increased Windows Compatibility

rmkraus:
Windows protobuf descriptors don’t seem to accept the syntax argument.
Removing it seems to cause to issues. This update will remove the
syntax argument if it is not present in the FileDescriptor. It will
assume that if the syntax argument is not in one, then it is not any of
the Descriptors.


rmkraus:
This PR is to address balloob/home-assistant#221.

balloob:
Were you running [Protobuf 3.0.0a3](https://pypi.python.org/pypi/protobuf/3.0.0a3) ? 

The file you adjusted has been automatically generated by 3.0.0a3 and should be compatible with it.

rmkraus:
Please do not power off or unplug your machine. Installing update....

I'll let you know whenever my Windows machine comes back online. I want to say it was 3.0.0a1. It might be worth noting that this is the version that installed when I ran the requirements.txt from home-assistant through pip. I haven't traced through to figure out why.

rmkraus:
pip install --upgrade pychromecast
and everything works correctly. So the real problem is that the incorrect version of pychromecast was installed by HA. Looks like I barked down the wrong tree on this one. My bad.

balloob:
[You're not alone](https://github.com/balloob/pychromecast/issues/47#issuecomment-107822162)

rmkraus:
Ahh. I see. Maybe there should be a pip requirement for pychromecast then?

balloob:
Well [setup.py](https://github.com/balloob/pychromecast/blob/master/setup.py) should ensure that the dependencies get installed when installing PyChromecast. Somehow older versions of pip interpret the versions wrong.

rmkraus:
Sorry, I wasn't clear. I meant to put install_requires=['pip>=xxx', ....] to setup.py. Although that won't work because it would update pip for the next run (if it even works at all). So it may be worth noting that this patch would fix pychromecast when it is installed with older version of pip. So that's something.

balloob:
It's in the [PyChromecast README](https://github.com/balloob/pychromecast#dependencies) but HA users won't read it

rmkraus:
haha, yeah, case in point

PR #56, opened 2015-08-23
Updated socket client

rmkraus:
The socket client was updated to eliminate nuisance errors. Before,
errors were regularly thrown that the socket connection was dropped and
the pychromecast would be reconnecting. This no longer happens.

An error also existed where Chromecast connections were dropped after a
few hours. It is my guess that this commit will fix this issue, but
that has not yet been confirmed.

The PING command is also now sent to the Chromecast and the PONG
response is now handled. Pychromecast now uses this mechanism to
determine if a socket connection is dropped. PINGs are set every 10
seconds and a PONG is expected back within 10 seconds. The Chromecast
will usually automatically PING the client, but not always. Because of
this, the PINGs sent to the client were not relied upon.

All hardcoded times were brought to the top and put in constants. The
timeout time was reduced to 30 seconds from 60.


rmkraus:
By the way, don't mere yet. I'm still running some tests to make sure everything is working.

pychromecast/socket_client.py
```diff
@@ -442,13 +466,30 @@ def receive_message(self, message, data):
...
 
+    def reset(self):
+        """ Reset expired counter. """
+        self.last_pong = time.time()
+
+    @property
+    def expired(self):
+        """ Indicates if connection has expired. """
+        if time.time() - self.last_ping > HB_PING_TIME:
+            self.ping()
```
balloob: Properties should be read only and have no side effects.

balloob:
Looks good to me!

rmkraus:
This library passed a 12 hour test. So far it seems much more stable. I'm merging. Feel free to publish when you can.

Issue #40, opened 2015-03-13
Sleep required between calls


geuben:
This works:
```python
import pychromecast
import time
cast = pychromecast.get_chromecast()
print cast.device
time.sleep(1)
print cast.status
```

This doesn't, show status of None:
```python
import pychromecast
import time
cast = pychromecast.get_chromecast()
print cast.device
print cast.status
```

balloob:
That's exactly how it is done in [example.py](https://github.com/balloob/pychromecast/blob/master/example.py#L22) :+1:

geuben:
Which is how I figured out how to get it to work. Doesn't make it right though.

Do you know why the sleep is required? Would be nice to at least document its requirement.

balloob:
PyChromecast communicates with Chromecasts using a socket. When you initialize an object it will start the connection. It takes a bit of time for the Chromecast to report its initial status back to PyChromecast.

geuben:
I think it would be better for the library to block until the connection is established and the status is available. You could make an optional non_blocking call too incase there are people that wan't to use it in that way.

PR #63, opened 2015-10-26
Added wait() method to make it easier to wait for the initial status

am0s:
This removes the need to make a custom status listener that waits for
the initial status message. Once the status message has been received
the method will return, if a status is already present it returns
immediately.

The caller must do a:
```python
  cast = pychromecast.get_chromecast()
  cast.wait()
```
right after receiving the cast object.

This could help out with solving issue #40.

Open for a better name for the method.

### DIFF
--- a/pychromecast/__init__.py
+++ b/pychromecast/__init__.py
@@ -159,6 +159,7 @@ class Chromecast(object):
                 "Could not connect to {}".format(self.host))
 
         self.status = None
+        self.status_event = threading.Event()
 
         self.socket_client = socket_client.SocketClient(host, tries)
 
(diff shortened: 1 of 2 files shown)


pychromecast/__init__.py
```diff
@@ -237,6 +237,20 @@ def volume_down(self):
...
+        is ready as soon a status message has been received.
+
+        If the status has already been received then the method returns
+        immediately.
+
+        :param timeout: a floating point number specifying a timeout for the
+                        operation in seconds (or fractions thereof). Or None
+                        to block forever.
+        """
+        self.socket_client.receiver_controller.wait_for_status(timeout=timeout)
```
balloob: We should keep this logic in the Cast class. We can have the event be set when the cast class receives a new status

am0s:
The code has now been moved to the Cast class.

PR #64, opened 2015-10-28
Improved handling of connection

am0s:
Misc improvements to the socket connection handling.

* Avoids inifinite connection loops if the client was told to disconnect.
* Allows a caller to listen for connection changes, e.g. to listen for a disconnection.
* The time between connection retries can be specified by the caller.

### DIFF
--- a/pychromecast/socket_client.py
+++ b/pychromecast/socket_client.py
@@ -248,6 +248,14 @@ class SocketClient(threading.Thread):
         return self._request_id
 
     @property
+    def is_connected(self):
+        """
+        Returns True if the client is connected, False if it is stopped
+        (or trying to connect).
+        """
+        return not self.connecting
+
+    @property
     def is_stopped(self):
         """
         Returns True if the connection has been stopped, False if it is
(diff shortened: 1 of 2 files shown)


pychromecast/socket_client.py
```diff
@@ -809,4 +891,10 @@ def tear_down(self):
         """ Called when controller is destroyed. """
         super(ReceiverController, self).tear_down()
 
+        self.status = None
+        self.launch_failure = None
+        self.app_to_launch = None
```
balloob: These variables seem not to be used.

balloob: Ah, I see now that not all code is included in this PR but some have been committed. Nevermind.

balloob:
It looks good. Awesome.

If you can give me your PyPi username, I will add you as an owner so you can make releases too.

pychromecast/socket_client.py
am0s: Ah yes, the previous commits were meant as a pull request as well but I misunderstood the UI in sourcetree and it got pushed directly.

PR #69, opened 2016-01-03
Increasing socket reliability

rmkraus:
This PR intends to increase the reliability of the socket connection. At the moment it drops out after a few hours. Changes are being made in this PR to fix the Exceptions that are causing the socket to drop out. The following changes have been made so far:

1. Separated the logged error from a heartbeat timeout and socket communication error for easier debugging.
2. Changed the error being raised when reading from a broken socket from RuntimeError to socket.error. This is the error that the run loop of the thread expects when interfacing with the socket.
3. Reset the force_recon flag immediately after a successful reconnection. At the moment, an Exception is sometimes raised when updating the status after a forced reconnection because the class assumed the socket was still broken.
4. Prevented a Ping command from being sent if the socket is broken.
5. Handle NotConnected exceptions when sending a Ping command.

~~This PR is not ready to be merged.~~


rmkraus:
Update: With these updates, I have about 24 hours with a persistant connection. I would like to go at least 48 hours without having to restart the Cast class before publishing. 

As an aside, the socket drops out constantly. There is an awful lot of log file spamming under normal operation. This should probably be addressed as well. Just not in this PR.

balloob:
Awesome sauce @rmkraus !

rmkraus:
Ran for 48 hours without permanently losing connection. :+1:

Ready for review/merge.

I'll have another PR to address the logging levels of a lot of the socket messages.

Issue #60, opened 2015-09-23
Log spamming.


stefan-jonasson:
I get these in my logs every 5th second:

```
 ERROR (Thread-16) [pychromecast.socket_client] Failed to connect, retrying in 5s
Traceback (most recent call last):
  File "/home/stefan/.homeassistant/lib/pychromecast/socket_client.py", line 160, in initialize_connection
    self.socket.connect((self.host, 8009))
  File "/usr/lib/python3.4/ssl.py", line 846, in connect
    self._real_connect(addr, False)
  File "/usr/lib/python3.4/ssl.py", line 837, in _real_connect
    self.do_handshake()
  File "/usr/lib/python3.4/ssl.py", line 810, in do_handshake
    self._sslobj.do_handshake()
ConnectionResetError: [Errno 104] Connection reset by peer
```

I think it would be enough with an INFO level log instead?

balloob:
Well no, it's an error that shouldn't happen. It's your chromecast hanging
up on us.

stefan-jonasson:
I don't want to unplug my chromecast because it works. As I understand it I cannot disable the chromecast detection in the Home assistant discovery module so there is no way for me to stop this logging. Maybe we could add some try x times and then wait for some longer period and try again. 

Maybe just remove the traceback so it's not so verbose! 
If this is happening for me it's probably happening for other chromecasts as well.

balloob:
I have no issues here. There is some max retries code available in pychromecast but we're not using it right now within Home Assistant.

As a quick solution for you: write a custom component that disables error logging for pychromecast

balloob:
The traceback is helpful in debugging if any other error is happening. It is an error after all.

For HA, you could write a custom component to disable pychromecast error logging.

ahocquet:
I had the same logs tonigh in my daemon script used to monitor chromecast. The problem is that after I got this "spam" logs, pychromecast is unable to do anything with my chromecast. I need to force quit my script and relaunch it.

Here's the logs : 

```
2015-12-14 03:29:58,579 - pychromecast.controllers - INFO - Receiver:Stopping current app 'E8C28D3C'
2015-12-14 04:37:01,222 - pychromecast.socket_client - ERROR - Error communicating with socket, resetting connection
2015-12-14 04:37:09,217 - pychromecast.socket_client - ERROR - Failed to connect, retrying in 5.000000s
Traceback (most recent call last):
[...]

PR #71, opened 2016-01-09
Demoted many log levels in socket_client

rmkraus:
1. Changed logging for repeated connection failures. First attempt is
logged as EXCEPTION, the rest are logged as DEBUG. If no connection is
made and tries expire, an error is still logged at ERROR. This addresses issue #60.
2. All issues that cause the socket to be reset (read/write errors) are
now logged at INFO and the general message was demoted to WARNING as
the IO errors are handled exceptions that will not generally cause the
application to crash.
3. Unhandled messages to defined namespaces have been demoted to DEBUG.
This was done to match the logging level used to indicate an unhandled
namespace.
4. Status updates have been demoted to DEBUG as they might be too
verbose for INFO.

### DIFF
--- a/pychromecast/socket_client.py
+++ b/pychromecast/socket_client.py
@@ -202,6 +202,7 @@ class SocketClient(threading.Thread):
         self._open_channels = []
 
         self.connecting = True
+        retry_log_fun = self.logger.exception
 
         while not self.stop.is_set() and (tries is None or tries > 0):
             try:
@@ -231,8 +232,10 @@ class SocketClient(threading.Thread):
 
                 self._report_connection_status(
                     ConnectionStatus(CONNECTION_STATUS_FAILED, None))
-                self.logger.exception("Failed to connect, retrying in %fs",
-                                      self.retry_wait)
+                retry_log_fun("Failed to connect, retrying in %fs",
+                              self.retry_wait)
+                retry_log_fun = self.logger.debug
+
                 time.sleep(self.retry_wait)
                 if tries:
                     tries -= 1
@@ -330,7 +333,7 @@ class SocketClient(threading.Thread):
                     raise
                 except socket.error:
                     self._force_recon = True
-                    self.logger.error('Error reading from socket.')
+                    self.logger.info('Error reading from socket.')
                 else:
                     data = _json_from_message(message)
             if not message:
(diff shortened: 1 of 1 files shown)


pychromecast/socket_client.py
```diff
@@ -231,8 +232,10 @@ def initialize_connection(self):
 
                 self._report_connection_status(
                     ConnectionStatus(CONNECTION_STATUS_FAILED, None))
-                self.logger.exception("Failed to connect, retrying in %fs",
-                                      self.retry_wait)
+                retry_log_fun("Failed to connect, retrying in %fs",
+                              self.retry_wait)
+                retry_log_fun = self.logger.debug
```
balloob: Could there be a moment you want to reset this one to exception?

rmkraus: The way I have it now, an error will always be thrown on the first failed
connection attempt. Every time pychromecast has to reconnect, this will be
reset.

I think this should be ok. In regular use, if pychromecast isn't working,
there will be some messages letting you know that it dropped a connection
and had trouble reconnecting. If someone were to have a lot of issues, they
should log info and debug messages to get more detail on what is happening.

Now that I'm thinking about this again, though. Maybe this should log to
info rather than debug after repeating failed retries. If you'd prefer, I
can also elevate it back to exception after x failed retries. I didn't do
that because of the reasons discussed in #60 with users turning their Cast
off when their tv is off.

On Saturday, January 9, 2016, Paulus Schoutsen <notifications@github.com>
wrote:

> In pychromecast/socket_client.py
> <https://github.com/balloob/pychromecast/pull/71#discussion_r49261442>:
>
> > @@ -231,8 +232,10 @@ def initialize_connection(self):
> >
> >                  self._report_connection_status(
> >                      ConnectionStatus(CONNECTION_STATUS_FAILED, None))
> > -                self.logger.exception("Failed to connect, retrying in %fs",
> > -                                      self.retry_wait)
> > +                retry_log_fun("Failed to connect, retrying in %fs",
> > +                              self.retry_wait)
> > +                retry_log_fun = self.logger.debug
>
> Could there be a moment you want to reset this one to exception?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/balloob/pychromecast/pull/71/files#r49261442>.
>


-- 
Thanks,

Ryan Kraus

balloob: Nah I think it's fine. DIdn't realize it was inside the `initalize_connection` method.

balloob:
Code looks good :dolphin: 

You can merge when you feel it's ready.

rmkraus:
:+1: Merging now. I have no other PR's that I'm looking to do. If you're comfortable publishing now to PyPI, we can get this rolled into the next release of HA.

balloob:
Will do

PR #75, opened 2016-01-22
Feature/multiroom

am0s:
Adds support for multi-room casting as well as improved details about Audio cast devices.
The main thing to support multi-room is to use the port value reported by the cast device.

### DIFF
--- a/pychromecast/__init__.py
+++ b/pychromecast/__init__.py
@@ -13,7 +13,8 @@ from .config import *  # noqa
 from .error import *  # noqa
 from . import socket_client
 from .discovery import discover_chromecasts
-from .dial import get_device_status, reboot, DeviceStatus
+from .dial import get_device_status, reboot, DeviceStatus, CAST_TYPES, \
+    CAST_TYPE_CHROMECAST
 from .controllers.media import STREAM_TYPE_BUFFERED  # noqa
 
 IDLE_APP_ID = 'E8C28D3C'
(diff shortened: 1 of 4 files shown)


pychromecast/__init__.py
```diff
@@ -161,31 +178,72 @@ class Chromecast(object):
...
+                                  dev_status.manufacturer),
+                    api_version=(device.api_version or
+                                 dev_status.api_version),
+                    uuid=(device.uuid or
+                          dev_status.uuid),
+                    cast_type=(device.cast_type or
+                               dev_status.cast_type),
+                )
+            else:
+                self.device = DeviceStatus(
```
balloob: This seems like a one on one copy, can't we just omit it ?

balloob:
Looks very solid, nice job! :dolphin: 

Just one minor comment but this looks good to go.

pychromecast/__init__.py
am0s: It merges the two objects as we don't get all information from the dial event.
The remaining parts are filled in from the device description.
The dialog information is the most correct but it does not have manufacturer I believe it was.

But I'll see if I can rewrite this code to be a bit more elegant + a comment about this.

balloob: I see the merging in the above statement but isn't the one below only using data from `device` ?

am0s: Ah, I did that to get a copy but since these are tuples there's no point in that. I'll fix.

am0s:
Ok, updated the code for DeviceStatus, and also added a version variable which I noticed was missing

balloob:
Ok it looks all good. You can merge when you're ready. Want to push out a new version too?

am0s:
Sure, would this be a 0.7 release?

balloob:
I would imagine so, pretty significant feature bump

PR #77, opened 2016-02-16
Add metadata support to default media receiver

kharts:
I added extra parameter metadata to play_media() method of MediaController.

Example of usage:
```python
import pychromecast
cast = pychromecast.get_chromecast()
metadata = {
    "metadataType": 0,
    "title": "Big Buck Bunny",
    "images": [
        {"url": "http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/images/BigBuckBunny.jpg"}
    ]
}
cast.media_controller.play_media(
    'http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4',
    'video/mp4',
    metadata=metadata
)
```
Here is a full doc [https://developers.google.com/cast/docs/reference/messages#MediaData](https://developers.google.com/cast/docs/reference/messages#MediaData)

Hope it will be useful for someone.

### DIFF
--- a/pychromecast/controllers/media.py
+++ b/pychromecast/controllers/media.py
@@ -396,8 +396,27 @@ class MediaController(BaseController):
                    current_time=0, autoplay=True,
                    stream_type=STREAM_TYPE_BUFFERED,
                    metadata=None):
-        """ Plays media on the Chromecast. Start default media receiver if not
-            already started. """
+        """
+        Plays media on the Chromecast. Start default media receiver if not
+        already started.
+
+        Parameters:
+        url: str - url of the media.
+        content_type: str - mime type. Example: 'video/mp4'.
+        title: str - title of the media.
+        thumb: str - thumbnail image url.
+        current_time: float - seconds from the beginning of the media
+            to start playback.
+        autoplay: bool - whether the media will automatically play.
+        stream_type: str - describes the type of media artifact as one of the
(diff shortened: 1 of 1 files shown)


am0s:
Looks nice.

Could you document the new parameter and add a link to the google cast doc in the docblock?

kharts:
I had a courage to document all parameters. I think it will look more consistent.

balloob:
I like it :dolphin:

pychromecast/controllers/media.py
```diff
@@ -394,9 +394,29 @@ def _fire_status_changed(self):
...
-        """ Plays media on the Chromecast. Start default media receiver if not
-            already started. """
+                   stream_type=STREAM_TYPE_BUFFERED,
+                   metadata=None):
+        """
+        Plays media on the Chromecast. Start default media receiver if not
+        already started.
+
+        Parameters:
+        url: str - url of the media.
```
am0s: Great stuff.

Here's a small tip for documenting parameters in Python, if you use the restructuredtext syntax it will be picked up by the Sphinx doc generator and it is also supported by IDE's (such as PyCharm).
http://www.sphinx-doc.org/en/stable/domains.html#info-field-lists

In this case it would be:
```
:param url: url of the media
:type url: str
```

PR #82, opened 2016-03-22
Fix / Add support for metadata, title, and thumb

noob-4-life:


noob-4-life:
I'm not really sure about the **customdata['payload']** stuff. But this is what I've been using and it seems to work fine.

noob-4-life:
Sorry for this second comment. I forgot about python 3 support (dict.iteritems in python 2 is dict.items in python 3). But I'm not sure of the best way to handle this either.

PR #101, opened 2016-07-27
[WIP] DS Audio controller

wind-rider:
This is a pull request to add a controller for [DS Audio](https://dsaudio-chromecast.synology.com/dsaudio.html) - the audio app for Synology NASes.

The basic version is working; I tried to keep the interface as close as possible to the default media controller. Also I tried to keep the DsAudioMediaStatus class as close to the original MediaStatus class.

Refactoring work:
I still think that somehow pychromecast should load the appropriate controller for app that is running on the chromecast, with hopefully a common interface to talk to (like play(), pause() and seek()).

Also, `pychromecast.get_chromecast().media_controller` is kind of useless when another app is loaded - this could be improved if there would be the right controller for the current app.

@balloob could you help me with this refactoring work? (To give some guidance where to go and to give some advice how to implement it

Besides, I still want to add the album art.
- [ ] Refactoring
- [ ] linting
- [ ] album art

### DIFF
--- a/pychromecast/config.py
+++ b/pychromecast/config.py
@@ -9,6 +9,7 @@ APP_BACKDROP = "E8C28D3C"
 APP_YOUTUBE = "YouTube"
 APP_MEDIA_RECEIVER = "CC1AD845"
 APP_PLEX = "06ee44ee-e7e3-4249-83b6-f5d0b6f07f34_1"
+APP_DS_AUDIO = "ED01B6D7"
 
 
 def get_possible_app_ids():
--- a/pychromecast/controllers/ds_audio.py
+++ b/pychromecast/controllers/ds_audio.py
@@ -1,0 +1,715 @@
+"""
+Controller to interface with the DS Audio-app at https://dsaudio-chromecast.synology.com/dsaudio.html
+"""
+from collections import namedtuple
+
+from ..config import APP_DS_AUDIO
+from . import BaseController
+
+# STREAM_TYPE_UNKNOWN = "UNKNOWN"
+# STREAM_TYPE_BUFFERED = "BUFFERED"
+# STREAM_TYPE_LIVE = "LIFE"
+
+AUDIO_PLAYER_REPEAT_NONE = "none"
+AUDIO_PLAYER_REPEAT_ONE = "one"
+AUDIO_PLAYER_REPEAT_ALL = "all"
+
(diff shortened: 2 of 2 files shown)


balloob:
Let's not include refactoring as part of this work and just get the controller app to work. One thing per PR 👍 

We can have some app id to controller mapping things, but if you want to include it in something like Home Assistant, the media controller is the way to go because it's 1 interface to rule them all.

Does the DS Audio controller offer anything beyond the normal media controller app?

wind-rider:
@balloob I don't understand why you want 1 media controller. The issue is that some chromecast apps do not talk the protocol of the "Default media receiver app".

So for play/pause in ds audio I need pychromecast to send different messages than to the default app.

Besides it needs an authentication in the "connected"

Also the "status update messages" are different from the MediaStatus messages (although there is some overlap)

I would say that controllers should preferably have a common interface (all controllers should have a play method, a pause method, etc) with feature detection which features are supported

The appropriate controller for the current app could then be loaded in a property of the pychromecast client.

HomeAssistant should then be able to call play or pause without knowing what the implementation is. Simply using the default media controller for all apps does not work...

wind-rider:
There are some methods and info that the default controller does not offer, like a repeat mode, shuffle mode and selecting a song from a playlist.

Also the ds audio app sends the playlist to the controller when the connection is made

balloob:
So what you are talking about is a media controller interface that is not the actual media controller app, which can be implemented by different controllers for different apps ?

wind-rider:
Yes, that's what I mean.

But I'm not familiar with Python - I'm a C# programmer.

I would for example make the BaseController abstract, with methods that all inheriting controllers have to implement (you already did this).

The BaseController is not meant to be used directly but it can contain shared logic and the events like `channel_connected`.

Since all methods are defined on the abstract BaseController, applications like HomeAssistant do not need to know whether they talk to BaseController, MediaController or DsAudioController.

As long as the right controller is provided by the `Chromecast` object, applications can call something like `pause()` and the loaded controller would then send the needed message.

wind-rider:
PS1: The difficult thing of finding a common interface (set of methods and properties that controllers would have to support) for controllers is that e.g. the methods to load new items are almost always different.

PS2: It seems that some things are set on Chromecast level (things like volume) and some things are set on application level (methods like play/pause and info about media items). Only the things on application level need to be made specific per controller/application

PR #104, opened 2016-08-21
Specify volume delta.

smblott-github:
Allow the user to specify the delta value for `volume_up()` and `volume_down()`.

(I find the hardwired delta of `0.1` too coarse for my own set up.)

### DIFF
--- a/pychromecast/__init__.py
+++ b/pychromecast/__init__.py
@@ -349,20 +349,20 @@ class Chromecast(object):
         """ Reboots the Chromecast. """
         reboot(self.host)
 
-    def volume_up(self):
-        """ Increment volume by 0.1 unless it is already maxed.
+    def volume_up(self, delta=0.1):
+        """ Increment volume by 0.1 (or delta) unless it is already maxed.
         Returns the new volume.
 
         """
-        volume = round(self.status.volume_level, 1)
-        return self.set_volume(volume + 0.1)
+        volume = round(self.status.volume_level, 2)
+        return self.set_volume(volume + delta)
 
-    def volume_down(self):
-        """ Decrement the volume by 0.1 unless it is already 0.
+    def volume_down(self, delta=0.1):
+        """ Decrement the volume by 0.1 (or delta) unless it is already 0.
         Returns the new volume.
         """
-        volume = round(self.status.volume_level, 1)
-        return self.set_volume(volume - 0.1)
+        volume = round(self.status.volume_level, 2)
+        return self.set_volume(volume - delta)
 
     def wait(self, timeout=None):
         """
--- a/pychromecast/__init__.py
+++ b/pychromecast/__init__.py
@@ -354,13 +354,17 @@ class Chromecast(object):
         Returns the new volume.
 
         """
-        return self.set_volume(self.status.volume_level + abs(delta))
+        if delta <= 0:
+            raise ValueError("volume delta must be greater than zero, not {}".format(delta))
+        return self.set_volume(self.status.volume_level + delta)
 
     def volume_down(self, delta=0.1):
         """ Decrement the volume by 0.1 (or delta) unless it is already 0.
         Returns the new volume.
         """
-        return self.set_volume(self.status.volume_level - abs(delta))
+        if delta <= 0:
+            raise ValueError("volume delta must be greater than zero, not {}".format(delta))
+        return self.set_volume(self.status.volume_level - delta)
 
     def wait(self, timeout=None):
         """
--- a/pychromecast/__init__.py
+++ b/pychromecast/__init__.py
@@ -360,7 +360,6 @@ class Chromecast(object):
         """ Decrement the volume by 0.1 (or delta) unless it is already 0.
         Returns the new volume.
         """
-        import sys
         return self.set_volume(self.status.volume_level - abs(delta))
 
     def wait(self, timeout=None):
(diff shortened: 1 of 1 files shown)


smblott-github:
The tests seem to be failing; however, I can't see where this PR is causing that.  Is there just a problem with the tests? (Or should I worry about this?)

pychromecast/__init__.py
```diff
@@ -349,20 +349,20 @@ def reboot(self):
...
-    def volume_up(self):
-        """ Increment volume by 0.1 unless it is already maxed.
+    def volume_up(self, delta=0.1):
+        """ Increment volume by 0.1 (or delta) unless it is already maxed.
         Returns the new volume.
 
         """
-        volume = round(self.status.volume_level, 1)
-        return self.set_volume(volume + 0.1)
+        volume = round(self.status.volume_level, 2)
```
balloob: You can't just  adjust the rounding to now match the delta that you're planning to use.

balloob:
It looks like flake8 got updated and this is causing our tests to fail. You can ignore any errors reported by the builds that are not related to your code.

pychromecast/__init__.py
smblott-github: 1. This PR is required (at least for me).  With a delta of `0.1`, it is not possible to select a reasonable volume level on one of my devices.  The "Volume Up" button on the device itself adjusts the volume by `0.02`, for example.

2. If you do rounding here, then you have to pick a number (`2` here).  Two significant digits seems a reasonable choice to me, three might be reasonable too. One digit is known to be inadequate, and four or more would be overkill.  So which number to pick?  `2` seems reasonable.

3. Why round at all?  It's not clear what the advantage of rounding is.  Why not just take it out and let the user decide.

4. There should probably be `Math.max(0.0, new_volume)` and `Math.min(1.0, new_volume)` in here somewhere too.

balloob:
This PR seems to have gone stale. Closing it. You can reopen it when you're ready to finish it.

smblott-github:
In what way is it stale?  I argued above for why this is the right thing to do, and you didn't respond.  The bottom line is:

- The current rounding to 1 digit is inadequate for some devices.
- Either pick a better default number of digits (2, here), or allow the user to pick the number (this PR), or get rid of rounding altogether.

(I would just get rid of the rounding, myself.  It's not clear why you should round at all.)

balloob:
screen shot 2016-12-12 at 11 04 34 pm

This is without rounding, the values are not very representable.

If you want to have no rounding, best to implement your own method:

```python
cast.set_volume(cast.status.volume_level + 0.05)
```

smblott-github:
What does "presentable" mean?  Who knows?  This is a library, who knows what it's used for?

Whoever uses this library will make their own decision about how to present the data.  It could be a graphical slider.  It could be a cron job where the data is never presented to the user.  Who knows?

Regardless, rounding to a level that is known to be inadequate for some devices cannot be correct.

(Although, I accept that there is a workaround.)

balloob:
I guess you're right, let's get rid of the rounding. If you remove all the rounding, we can merge this PR.

smblott-github:
Rounding removed, see https://github.com/balloob/pychromecast/pull/104/commits/918151c2278951762b8c021f8967b4b6404149e3.

I tried using `max()` and `min()` to make sure we don't go outside of the `0.0` to `1.0` range, but that doesn't seem to be necessary.

I added `abs()` to ensure that `volume_up()` really does get louder, and `volume_down()` really does get quieter.  Without `abs()`, the latter could be confusing if a user tries something like:

```python
cast.volume_down(-0.05)
```

pychromecast/__init__.py
```diff
@@ -349,20 +349,19 @@ def reboot(self):
...
-    def volume_up(self):
-        """ Increment volume by 0.1 unless it is already maxed.
+    def volume_up(self, delta=0.1):
+        """ Increment volume by 0.1 (or delta) unless it is already maxed.
         Returns the new volume.
 
         """
-        volume = round(self.status.volume_level, 1)
-        return self.set_volume(volume + 0.1)
+        return self.set_volume(self.status.volume_level + abs(delta))
```
balloob: I don't like the usage of `abs` as it might be confusing for the user. Let's raise a ValueError instead when delta <= 0

pychromecast/__init__.py
```diff
@@ -349,20 +349,19 @@ def reboot(self):
...
 
-    def volume_down(self):
-        """ Decrement the volume by 0.1 unless it is already 0.
+    def volume_down(self, delta=0.1):
+        """ Decrement the volume by 0.1 (or delta) unless it is already 0.
         Returns the new volume.
         """
-        volume = round(self.status.volume_level, 1)
-        return self.set_volume(volume - 0.1)
+        import sys
```
balloob: stale import

smblott-github: Whoops.  That was from testing.  Sorry about that (too early in the morning, here). It's fixed: https://github.com/balloob/pychromecast/pull/104/commits/909e485e622bedfd181dae618690209a273218c6.

pychromecast/__init__.py
balloob: Did you see this comment?

smblott-github: No I missed it.  Fixed now.

balloob:
OK to merge when linting errors fixed!

balloob:
💃 🐬

PR #108, opened 2016-08-22
Adds support for own app-id

### LINKED ISSUE #106: Styled Media Receiver?
I have tried to find out in the source code, but I was not able to see if it is possible to connect somehow to a [Styled Media Receiver](https://developers.google.com/cast/docs/styled_receiver). 
A styled media receiver does not have its own namespace (as far as I understand) and therefore the example provided in the README doesn't work unfortunately. 

There is a `supporting_app_id`, but as far as I understand it is only called if the current namespace is unsupported.

wa4557:
Fixes #106. Figured it out shortly after I asked :) This allows to define its own app_id for the generic media controller, which allows to use custom styled receiver applications.

### DIFF
--- a/pychromecast/controllers/media.py
+++ b/pychromecast/controllers/media.py
@@ -260,7 +260,7 @@ class MediaController(BaseController):
 
         self.media_session_id = 0
         self.status = MediaStatus()
-
+        self.app_id = APP_MEDIA_RECEIVER
         self._status_listeners = []
 
     def channel_connected(self):
@@ -423,7 +423,7 @@ class MediaController(BaseController):
         https://developers.google.com/cast/docs/reference/messages#MediaData
         """
 
-        self._socket_client.receiver_controller.launch_app(APP_MEDIA_RECEIVER)
+        self._socket_client.receiver_controller.launch_app(self.app_id)
 
         msg = {
             'media': {


balloob:
I don't get this PR.

If you just want to launch the styled media receiver app id, you don't need to change the media controller.

The media controller will be able to control any app that supports the media namespace , so no need to be able to allow overwriting it. This is just a fallback method.

wa4557:
Ok but how would I do that?

Let's say I have made a styled Media Receiver that has a speific app id (AAAAAA).

As Far as I Unterstand a `register_handler` is not the right way to do it, because it is for custom Media Händler that have their own namespace. (Which a styled Media Receiver does not have, because thats the point of a styled Media Receiver)

So how would I launch my own app ID without changing the namespace? (I don't just want to Control it, I want to launch URLs on it)

Looking though the Code I can also see, that the function `play_media` will always launch the Default Media Receiver (if app_id != self.app_id in socket_client) so even if I can control my app I can not launch new URLs on it.

Plus it works the way I do it.

balloob:
You can just use the same call that this controller uses:

```python
receiver_controller.launch_app(app_id)
```

wa4557:
Fine but then I cannot use the DefaultMediaReceiver but have to write my own, right?

balloob:
I see what you mean now. The styled media player and the normal media player have the same control interface. The status updates will work regardless but you're interested in the play_media command. 

I think the more correct approach to solving this is to only launch the default media player if the current app id is not equal to default/styled media player.

wa4557:
Yeah, I guess that's precisely the point of a StyledMediaReceiver. Everything the same but a different app_id. 

The problem is that you are doing a `if self.app_id == app_id` in `socket-client` which will launch the DefaultmediaController each time I'm trying to load a new URL. So actually my fix does exactly what you are proposing. Changing this variable the function `launch_app` only does something if the variable changes again (which makes sense imo)

pychromecast/controllers/media.py
```diff
@@ -423,7 +423,7 @@ def play_media(self, url, content_type, title=None, thumb=None,
         https://developers.google.com/cast/docs/reference/messages#MediaData
         """
 
-        self._socket_client.receiver_controller.launch_app(APP_MEDIA_RECEIVER)
+        self._socket_client.receiver_controller.launch_app(self.app_id)
```
balloob: Wrap this line in an if statement:

```python
if self._socket_client.app_id not in (APP_MEDIA_RECEIVER, APP_STYLED_MEDIA_RECEIVER):
    self._socket_client.receiver_controller.launch_app(APP_MEDIA_RECEIVER)
```

balloob: I thought the styled media player was always the same ?

wa4557: Now you have to write your own, and Register it with Google

balloob:
My apologies, you were right. Merging this 👍

wa4557:
No worries. Thanks for this great piece of Software!

PR #114, opened 2016-09-15
Ignore benign error in destructor. Fix #113

### LINKED ISSUE #113: 'Chromecast' object has no attribute 'socket_client' when Chromecast not reachable
Would be good for an exception to be thrown if __init__ cannot connect, this isn't very graceful.

noob-4-life:
It’s not really an issue, but when the Chromecast object gets garbage
collected after failing to connect to a device (before creating a
SocketClient instance), its __del__ is called, which in turn calls
`self.socket_client.stop.set()`. This leads to the error described in
the mentioned issue being logged/printed.

### DIFF
--- a/pychromecast/__init__.py
+++ b/pychromecast/__init__.py
@@ -404,7 +404,10 @@ class Chromecast(object):
         self.socket_client.join(timeout=timeout)
 
     def __del__(self):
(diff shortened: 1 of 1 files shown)


balloob:
Thanks!

PR #121, opened 2016-10-12
Add buffering state to media player

jaharkes:
During playback, my chromecast fairly frequently reports a buffering state
although playback is uninterrupted. Because this state was previously unknown,
home-assistant would report an "off" state for the device.

This fills up the logbook/history with many unnecessary state transitions.
Treating buffering as playing avoids these transitions.

### DIFF
--- a/pychromecast/__init__.py
+++ b/pychromecast/__init__.py
@@ -22,7 +22,7 @@ __all__ = (
     'get_chromecasts', 'get_chromecasts_as_dict', 'get_chromecast',
     'Chromecast'
 )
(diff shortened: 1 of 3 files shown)


balloob:
Jan, long time no see 👋 

Ok to merge once linting issue resolved. Can you include a version bump too? I'll release right after.

jaharkes:
Yeah, had a pleasantly busy summer. Good to see how much progress you guys have made.

I found one version in `setup.py` and a second one that was out of sync in `pychromecast/__init__.py`. I updated both of them.

PR #124, opened 2016-10-20
Catch more general exception

molobrakos:
catch more generic ValueError since different json implementations seems to raise either ValueError, json.JSONDecodeError or json.decoder.JSONDecodeError


balloob:
Ah yes, json.decoder.JSONDecodeError extends ValueError and is new in 3.5, it should not have been used.

PR #125, opened 2016-10-24
Changing seek auto-play to option

Owlz:
I have a situation where i'd like to seek and have it paused. Adding this as an option to media.py but keeping the default as auto-play.


balloob:
Ok to merge when linting errors fixed.

balloob:
This PR seems to have gone stale. Closing it. You can reopen it when you're ready to finish it.

PR #126, opened 2016-10-29
Improve subtitles

palaviv:
This PR improve the subtitles on play_media:

1. Subtitles are loaded automatically if subtitle are given.
2. The subtitles style is changed from default to no background and black edge.


balloob:
Top! 🐬

PR #131, opened 2016-12-04
Support own main loop and non-blocking functions

dominikkarall:
Hi,

this PR allows the implementation of an own main loop in the application which uses pychromecast as a library.
Just call "run_once()" after you receive data on the socket instead of running a thread which waits for data. Furthermore it makes all calls non-blocking if you use "no_threading=1" argument. The only function which remains blocking is the search function for devices.

Please review my code and give me feedback. I'm sure that it's not perfect yet, but I would like to use this PR to get feedback from you and optimize the code till it can be implemented in the official repository.

Thanks,
Dominik

### DIFF
--- a/examples/blocking.py
+++ b/examples/blocking.py
@@ -1,5 +1,8 @@
 """
 Example that shows how the new Python 2 socket client can be used.
+
+Functions called in this example are blocking which means that
+the function doesn't return as long as no result was received.
 """
 
 from __future__ import print_function
--- a/examples/non_blocking.py
+++ b/examples/non_blocking.py
@@ -1,0 +1,82 @@
+"""
+Example that shows how the new Python 2 socket client can be used.
+
+All functions (except get_chromecast()) are non-blocking and
+return immediately without waiting for the result. You can use
+that functionality to include pychromecast into your main loop.
+"""
+
+from __future__ import print_function
+import time
+import select
+import sys
+import logging
+
+import pychromecast
+
+"""
+Put this code into your startup/init code.
+This one is the only blocking part.
+"""
+def initialize_chromecast():
+    print("Initialize Chromecast (blocking)...takes some time")
+    cast = pychromecast.get_chromecast(friendly_name="Wohnzimmer",
+                                       blocking=False)
+    return cast
+
+"""
+Check for cast.socket_client.get_socket() and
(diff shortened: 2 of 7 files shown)


balloob:
I don't like it that this logic is interwoven with the other logic. That makes it way too confusing.

For this to be added you should:

 - decouple the data receiving part from the data processing part
 - allow the data receiving part to be swapped out
 - add a `run_once` data receiving part

dominikkarall:
Thanks for your feedback!

`run_once` does nothing more than receiving the message, routing it and calling the event callback function if there is one registered. So I'm not sure which parts you want to decouple? Do you want that data receiving is a completely separate function which is called by `run_once`?

E.g.
```
def run_once(self):
    # poll socket only if no_threading is used,
    # otherwise select should be done in own main loop
    if no_threading == False:
        poll_socket()
    read_message_from_socket()
    process_message_and_trigger_callbacks()
```

balloob:
Just gave the PR a better look and all the callbacks and broken up methods add way too much complexity. It might solve your use case but it will scare away potential future contributors because they will no longer be able to read and understand the source code.

dominikkarall:
What do you think about the latest changes I did in the PR? It's not breaking any functionality which was there before, it just let's the user decide if he wants to use the `blocking=0` version of pychromecast which would require to call `run_once()` by main loop.
If no `blocking` argument is added the functionality is exactly the same as it was before. So the user can decide if he want's to use
- blocking version: Mainly for scripts or other applications which integrate pychromecast in a separate process/thread
- non-blocking version: Users who have already their own main loop and don't want to have blocking functions (or new processes/threads to make them non-blocking) in place.

balloob:
Okay, you changed my mind :) I like it.

balloob:
Pylint is still reporting some errors. See https://travis-ci.org/balloob/pychromecast/jobs/181721687#L361-L365

balloob:
Could you add a new example directory with a non-blocking example integration?

dominikkarall:
- Fixed all Pylint errors
- Created a new examples directory
- Moved example.py to examples/blocking.py plus some comment
- Created a new non-blocking example in the examples directory

Please let me know if it is ok now and if you want to have it in a single commit.

balloob:
It's fine, GitHub has a squash function 👍

PR #134, opened 2016-12-06
Support for non-blocking callback-based discovery

armills:
Two new methods `start_discovery` and `stop_discovery` are added here. These allow non-blocking discovery of chromecasts.


balloob:
Nice! 🐬

PR #136, opened 2016-12-13
Non-blocking discovery in get_chromecasts

armills:
Integrate non-blocking discovery into get_chromecast methods.

It might make sense for these to be separate functions, but since we've included the blocking argument anyway, I went ahead with this implementation.

### DIFF
--- a/pychromecast/discovery.py
+++ b/pychromecast/discovery.py
@@ -90,7 +90,6 @@ def start_discovery(callback=None):
 
 def stop_discovery(browser):
     """Stop the chromecast discovery thread."""
-    browser.cancel()
     browser.zc.close()
 
 
--- a/examples/non_blocking.py
+++ b/examples/non_blocking.py
@@ -15,32 +15,32 @@ import logging
 import pychromecast
 
 """
-Put this code into your startup/init code.
-This one is the only blocking part.
-"""
-def initialize_chromecast():
-    print("Initialize Chromecast (blocking)...takes some time")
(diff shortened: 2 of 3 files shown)


pychromecast/discovery.py
```diff
@@ -90,7 +90,6 @@ def start_discovery(callback=None):
 
 def stop_discovery(browser):
     """Stop the chromecast discovery thread."""
-    browser.cancel()
```
armills: This was unnecessary since closing zeroconf also closes all listeners. Furthermore, this caused an exception if called in one of the callbacks, since it tried to join the discovery thread to itself.

balloob:
It looks good, ok to merge when lint issues resolved (see travis).

I wonder if it's time for a PyChromecast 1.0 release in which we will also clean up the discovery methods. Especially all the filter stuff for discovery is out of scope for the project.

armills:
At least on the non-blocking front, I know get_device_status is still performing IO on the main thread when a Chromecast object is initialized. I'd like to get that corrected first.

I agree the filtering makes more sense as downstream logic. I think get_chromecasts_as_dict might be worth removing as well for the same reasons.

https://github.com/balloob/pychromecast/blob/master/pychromecast/dial.py#L38

balloob:
I agree. They were contributions in the beginning of my open source career
when I wasn't that good at saying no yet

armills:
FYI, contrary to my last post, as long as downstream fetches Chromecasts with the `get_chromecasts` method, the blocking IO in the Chromecast `__init__` function happens in the `internal_callback` before calling the main callback, so it should be good.

PR #137, opened 2016-12-13
Update get_chromecasts_as_dict for non-blocking

armills:
Fix for get_chromecasts_as_dict with new non-blocking logic.


balloob:
Let's just remove it. I think having just get_chromecasts is enough

armills:
Opened https://github.com/balloob/pychromecast/pull/138

PR #138, opened 2016-12-13
Remove get_chromecasts_as_dict

armills:


balloob:
This is Good.

We should make sure to mention this in the release notes when we release v1

Issue #142, opened 2016-12-15
Protobuf requirement stuck at 3.0.0b2


theychx:
Hi
Protobuf is now at version 3.1.0. Is there any particular reason pychromecast is hardwired to version 3.0.0b2?

balloob:
Nop, PR welcome 👍

PR #143, opened 2016-12-16
Update protobuf requirement

theychx:
Here ya go (for #142)

### DIFF
--- a/chromecast_protobuf/authority_keys.proto
+++ b/chromecast_protobuf/authority_keys.proto
@@ -1,9 +1,13 @@
 // Copyright 2014 The Chromium Authors. All rights reserved.
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
+
 syntax = "proto2";
+
 option optimize_for = LITE_RUNTIME;
-package extensions.core_api.cast_channel.proto;
+
+package extensions.api.cast_channel.proto;
+
 message AuthorityKeys {
   message Key {
     required bytes fingerprint = 1;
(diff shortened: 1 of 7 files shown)


balloob:
We should require at least version 3.0 but no longer any beta's. We will also have to recompile the proto files to Python, in case some breaking changes sneaked in after beta 2.

theychx:
> We will also have to recompile the proto files to Python

I need some instructions with this

armills:
There are instructions here: https://developers.google.com/protocol-buffers/docs/pythontutorial#compiling-your-protocol-buffers

The three .proto files in chromecast_protobuf need to be compiled into the three *_pb2.py files in pychromecast.

theychx:
The proto files have been updated in the chromium git repo, since they were imported to [chromecast_protobuf](../tree/master/chromecast_protobuf). Should I update those before proceeding? Also, I plan using protoc 3.1.0 (the pre-compiled version from the google github repo). Let me know if you want me to use 3.0.0, or whatever.

armills:
Try updating to the latest proto files from chromium. It looks like they've only added new optional features, so nothing should break. Make sure you test the library as well as you can after the change.

It's probably safest to use protoc 3.0.0, but it shouldn't make a difference.

theychx:
I have run [/examples/non_blocking.py](../blob/master/examples/non_blocking.py) without problems and also done some casual `media_controller` tests myself. I am really not super familiar with pychromecast, so if you have some additional tests/examples you want me to run, I'll be happy to do that.

The files generated by `protoc 3.0.0` and `protoc 3.1.0` are identical.

balloob:
Awesome, thanks!

PR #145, opened 2016-12-21
Add media controller wait method

### LINKED ISSUE #128: .pause() Works in python cmd line but not in when running as script
Hi,

I am playing with this library and can't get pause to work in a script.
If I do:
```
from __future__ import print_function
import time
import pychromecast
cast = pychromecast.get_chromecast(friendly_name="Bedroom Chromecast")
# Wait for cast device to be ready
cast.wait()
print(cast.device)
print(cast.status)
mc = cast.media_controller
mc.pause()
time.sleep(5)
mc.play()
```

In a python cmd session, it works as expected. If I then put this into a .py file and run it, again it all works except for the mc.pause() part. There is no error output and the script appears to finish without issue.

Can anyone suggest what might be going on?

armills:
This PR adds a wait method to the media controller. This is convenient since commands to the media controller aren't allowed until the status has been received.

Fixes #128

### DIFF
--- a/pychromecast/controllers/media.py
+++ b/pychromecast/controllers/media.py
@@ -300,13 +300,16 @@ class MediaController(BaseController):
         self.send_message({MESSAGE_TYPE: TYPE_GET_STATUS},
                           callback_function=callback_function_param)
 
-    def _send_command(self, command):
+    def _send_command(self, command, blocking=True, timeout=None):
         """ Send a command to the Chromecast on media channel. """
         if self.status is None or self.status.media_session_id is None:
-            self.logger.warning(
-                "%s command requested but no session is active.",
-                command[MESSAGE_TYPE])
-            return
+            if blocking:
+                self.wait(timeout)
+            else:
+                self.logger.warning(
+                    "%s command requested but no session is active.",
+                    command[MESSAGE_TYPE])
+                return
 
         command['mediaSessionId'] = self.status.media_session_id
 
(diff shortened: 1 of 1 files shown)


palaviv:
I think this is a great enhancement. I currently use `update_status(blocking=True)` to do the same but this solution is better. I would like to suggest adding a `blocking` parameter to the `play_media` method that will call `wait` internally in `play_media`.

armills:
I like that idea. `play_media` doesn't require the status update, but I added the wait by default to the other commands. This should also fix #150.

pychromecast/controllers/media.py
```diff
@@ -297,10 +300,16 @@ def update_status(self, callback_function_param=False):
         self.send_message({MESSAGE_TYPE: TYPE_GET_STATUS},
                           callback_function=callback_function_param)
 
-    def _send_command(self, command):
+    def _send_command(self, command, blocking=True, timeout=None):
```
balloob: I don't like this to be added to every command, instead it is up to the developer using the lib to call a wait method.

pychromecast/controllers/media.py
```diff
@@ -361,31 +370,51 @@ def skip(self):
...
-            MESSAGE_TYPE: TYPE_EDIT_TRACKS_INFO,
-            "activeTrackIds": []
-        })
+        self._send_command(
+            {MESSAGE_TYPE: TYPE_EDIT_TRACKS_INFO,
+             "activeTrackIds": []},
+            blocking, timeout
+        )
+
+    def wait(self, timeout=None):
```
balloob: The name is not representative of what it does. `block_till_status_received` maybe? 

I wonder if we can achieve this by wrapping the status changed bit instead of adding a new threading event?

armills: To the second point, I think keeping the thread blocking to a dedicated method makes it more clear that the call will block. Wrapping the status might be unexpected.

pychromecast/controllers/media.py
```diff
@@ -381,11 +387,27 @@ def disable_subtitle(self):
...
+                        to block forever.
+        """
+        self.status_event.wait(timeout=timeout)
+
     def _process_media_status(self, data):
         """ Processes a STATUS message. """
         self.status.update(data)
 
         self.logger.debug("Media:Received status %s", data)
+        self.status_event.set()
```
balloob: Should we ever reset this ?

armills: Good thought. I've slightly reworked/renamed this to reflect the original goal of waiting until a media session is active and ready to accept commands. The event resets itself if the media session id is reset to None.

PR #158, opened 2017-01-24
Fix non-blocking initialize_connection

dominikkarall:
Fix non-blocking run_once() usage to solve issues if connection drops which would currently block for 30s.


armills:
This doesn't make it non-blocking, it makes it block for 0.1 seconds instead. Connecting to the socket is a blocking method. Basically you've just created new defaults for tries and timeouts. By default it already only tries once, and if you need a 0.1 second timeout you can pass that to the constructor.

dominikkarall:
You are right...that's not the proper solution. Let me think about real non-blocking the next days...

armills:
With the objects we're using to connect right now, the only option is to do the socket connection on a separate thread, although I don't know that that's something we want to try and do within this library. It might be better done by the user.

Issue #149, opened 2016-12-29
Add SSDP discovering


lars18th:
Hi devs,

At time, this library uses mDNS for discover Chromecast devices.
However, it exists another way for dicovering: SSDP.
See this tool also writen in Python: http://github.com/Pat-Carter/stream2chromecast

So, my suggestion is to add support for both mDNS and SSDP discovery.
You can use as example this code:
http://github.com/Pat-Carter/stream2chromecast/blob/master/cc_device_finder.py

Why this?
- mDNS doesn't work on all machines.
- mDNS is slow (in my Linux machine "avahi-browse --all" needs 60 seconds to discover "_goolecast._tcp" services).
- SSDP is quite simple!

Then, please, you like to consider this enhancement?

PR #161, opened 2017-02-03
[WIP] Add ssdp discovery

palaviv:
This PR add the option to discover chromecast devices using ssdp as requested in #149.

This is WIP as I am not sure if this is wanted. The ssdp discovery is simple but integrating it to the `get_chromecasts` proved to be more complex.

Few question I have:
1. Should I use `CastListener` and add optional ssdp and call it from my thread?
2. Is there a reason the we use `get_service_info` after getting the device name if we call `get_device_status` in `Chromecast` `__init__`?

### DIFF
--- a/pychromecast/__init__.py
+++ b/pychromecast/__init__.py
@@ -30,27 +30,31 @@ NON_UNICODE_REPR = sys.version_info < (3, )
 
 
 def _get_chromecast_from_host(host, tries=None, retry_wait=None, timeout=None,
-                              blocking=True):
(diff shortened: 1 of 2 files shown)


balloob:
We had SSDP in the past but i removed it because it was not giving us accurate results. With Cast groups one Chromecast can host multiple cast instances and that won't work with SSDP.

palaviv:
Cool. What about the second point @balloob?

balloob:
Sorry, I don't remember anymore.

PR #167, opened 2017-03-18
Dynamic host ip tracking for cast groups

### LINKED ISSUE #165: Cast groups changing "elected leader"
Hi,
I'm having some trouble with the chromecast audio group functionality in pychromecast, it seems like the cast group every now and then changes the "elected leader" of the group (and thereby the ip-adress of the group) but remains on the same port-number. Below is a log from HomeAssistant discovery showing the behaviour.

The first three lines shows the discovery of 3 "real" chromecast audio devices. Line 4 shows the discovery of a cast group on ip=192.168.0.205 and port 42675. Line 5 and 6 shows that after some (random?) time, the ip of the group changes (new device is discovered) but the port number stays the same. A Chromecast object associated with the first ip of the group will now return an error (see bottom) and a new Chromecast object (with the new ip) has to be created in order to connect.

Chromcast objects with "dynamic ip", is that possible? Maybe in line with the approach brought up in #111 ?

`17-03-12 12:55:46 INFO (MainThread) [homeassistant.components.discovery] Found new service: google_cast ('192.168.0.100', 8009)`

`17-03-12 12:55:46 INFO (MainThread) [homeassistant.components.discovery] Found new service: google_cast ('192.168.0.104', 8009)`

`17-03-12 12:55:46 INFO (MainThread) [homeassistant.components.discovery] Found new service: google_cast ('192.168.0.205', 8009)`

`17-03-12 12:55:46 INFO (MainThread) [homeassistant.components.discovery] Found new service: google_cast ('192.168.0.205', 42675)`

`17-03-12 13:28:29 INFO (MainThread) [homeassistant.components.discovery] Found new service: google_cast ('192.168.0.100', 42675)`

`17-03-12 15:17:08 INFO (MainThread) [homeassistant.components.discovery] Found new service: google_cast ('192.168.0.104', 42675)`

Error in homeassistant after ip of group has changed:
`  File "/home/hass/.homeassistant/deps/pychromecast/socket_client.py", line 589, in send_message
    raise NotConnected("Chromecast is connecting...")`

freol35241:
fixes #165

### DIFF
--- a/pychromecast/__init__.py
+++ b/pychromecast/__init__.py
@@ -198,7 +198,8 @@ class Chromecast(object):
             self.socket_client.start()
 
         # Set to True by default if cast_type == CAST_TYPE_GROUP
-        self.dynamic_host = True if self.cast_type == CAST_TYPE_GROUP else False
+        self.dynamic_host = True if self.cast_type == CAST_TYPE_GROUP \
+            else False
         self.dynamic_host = kwargs.pop('dynamic_host', self.dynamic_host)
         if self.dynamic_host:
             self.start_dynamic_host_tracking()
@@ -347,7 +348,8 @@ class Chromecast(object):
         """
         Start dynamic non-blocking host tracking
         """
-        self.tracker, self.browser = start_discovery(self._dynamic_host_callback)
+        self.tracker, self.browser = \
+            start_discovery(self._dynamic_host_callback)
 
     def _dynamic_host_callback(self, unit_id):
         """
@@ -359,15 +361,15 @@ class Chromecast(object):
         name = unit[4]
 
         if not name == self.name or not port == self.port:
-            return #Wrong group, keep discovering
+            return  # Wrong group, keep discovering
 
         if host == self.host:
-            return #No change in host, keep discovering
+            return  # No change in host, keep discovering
 
-        self.logger.info('Change in host ip detected, reinitializing object...')
+        self.logger.info('Change in host ip detected, '
+                         'reinitializing object...')
         stop_discovery(self.browser)
         self.__init__(host, port=port, device=self.device, dynamic_host=True)
-
 
     def __del__(self):
         try:
--- a/pychromecast/__init__.py
+++ b/pychromecast/__init__.py
@@ -32,6 +32,7 @@ NON_UNICODE_REPR = sys.version_info < (3, )
 CONNECTION_STATUS_CONNECTING = socket_client.CONNECTION_STATUS_CONNECTING
 CONNECTION_STATUS_FAILED = socket_client.CONNECTION_STATUS_FAILED
 CONNECTION_STATUS_CONNECTED = socket_client.CONNECTION_STATUS_CONNECTED
+CONNECTION_STATUS_DISCONNECTED = socket_client.CONNECTION_STATUS_DISCONNECTED
 
 
 def _get_chromecast_from_host(host, tries=None, retry_wait=None, timeout=None,
@@ -180,7 +181,6 @@ class Chromecast(object):
 
         self.browser = None
         self.tracker = None
-        self._socket_connection_status = None
 
         self.setup(host, port, self.device, self.kwargs)
 
@@ -188,7 +188,8 @@ class Chromecast(object):
         """
         Method for setting up, and resetting the chromecast object
         """
-        tries = kwargs.pop('tries', None)
+        tries = 2 if self.cast_type == CAST_TYPE_GROUP \
+            else kwargs.pop('tries', None)
         timeout = kwargs.pop('timeout', None)
         retry_wait = kwargs.pop('retry_wait', None)
         blocking = kwargs.pop('blocking', True)
@@ -217,11 +218,10 @@ class Chromecast(object):
             self.socket_client.start()
 
         # Set to True by default if cast_type == CAST_TYPE_GROUP
-        self.dynamic_host = True if self.cast_type == CAST_TYPE_GROUP \
-            else False
-        self.dynamic_host = kwargs.pop('dynamic_host', self.dynamic_host)
+        self.dynamic_host = kwargs.pop('dynamic_host', True if
+                                       self.cast_type == CAST_TYPE_GROUP
+                                       else False)
         if self.dynamic_host:
-            self._socket_connection_status = None
             self.start_dynamic_host_tracking()
 
     @property
@@ -374,38 +374,31 @@ class Chromecast(object):
         """
         Callback for socket client connection status
         """
-        if not status.address.address == self.host:
-            return  # Status message for previous socket_client
-
-        if status.status == CONNECTION_STATUS_CONNECTING:
+        if not status.status == CONNECTION_STATUS_DISCONNECTED:
             return  # Not interesting
 
-        if status.status == CONNECTION_STATUS_CONNECTED and self.browser:
+        def discovery_timeout():
+            """
+            Internal scheduled timeout for discovery, 5 seconds
+            """
+            time.sleep(10)
+            if not self.browser:
+                return  # Change in ip detected, discovery did not time out
+
+            self.logger.info('Discovery timed out, stopping')
             stop_discovery(self.browser)
             self.browser = None
-            self.logger.info('Reconnected, dynamic host discovery disrupted')
+            if not self.socket_client.is_connected:
+                # If no host on new ip address was found, try reconnecting
+                self.disconnect()
+                self.setup(self.host, self.port, self.device, self.kwargs)
 
-        if self._socket_connection_status == CONNECTION_STATUS_FAILED and \
-           status.status == CONNECTION_STATUS_FAILED and not self.browser:
-            # Connection and reconnection failed, see if reconnection can
-            # be made on another ip-adress
+        self.logger.info('Dynamic host discovery started')
+        self.tracker, self.browser = \
+            start_discovery(self._dynamic_host_callback)
 
-            def discovery_timeout():
-                """
-                Internal scheduled timeout for discovery, 5 seconds
-                """
-                time.sleep(5)
-                self.logger.info('Discovery timed out, stopping')
-                stop_discovery(self.browser)
-
-            self.logger.info('Dynamic host discovery started')
-            self.tracker, self.browser = \
-                start_discovery(self._dynamic_host_callback)
-
-            timeout = threading.Thread(target=discovery_timeout)
-            timeout.start()
-
-        self._socket_connection_status = status.status
+        timeout = threading.Thread(target=discovery_timeout)
+        timeout.start()
 
     def _dynamic_host_callback(self, unit_id):
         """
@@ -425,9 +418,13 @@ class Chromecast(object):
         self.logger.info('Change in host ip detected, '
                          'reinitializing object...')
         stop_discovery(self.browser)
+        self.browser = None
         self.disconnect()
         self.host = host
-        self.setup(host, port, self.device, self.kwargs)
+        try:
+            self.setup(host, port, self.device, self.kwargs)
+        except ChromecastConnectionError:  # noqa
+            pass  # DISCONNECTED status already sent, new discovery started
 
     def __del__(self):
         try:
--- a/pychromecast/__init__.py
+++ b/pychromecast/__init__.py
@@ -14,7 +14,7 @@ from .error import *  # noqa
 from . import socket_client
 from .discovery import discover_chromecasts, start_discovery, stop_discovery
 from .dial import get_device_status, reboot, DeviceStatus, CAST_TYPES, \
-    CAST_TYPE_CHROMECAST
+    CAST_TYPE_CHROMECAST, CAST_TYPE_GROUP
 from .controllers.media import STREAM_TYPE_BUFFERED  # noqa
 
 __all__ = (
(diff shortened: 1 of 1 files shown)


pychromecast/__init__.py
```diff
@@ -194,6 +197,13 @@ def __init__(self, host, port=None, device=None, **kwargs):
         if blocking:
             self.socket_client.start()
 
+        # Set to True by default if cast_type == CAST_TYPE_GROUP
+        self.dynamic_host = True if self.cast_type == CAST_TYPE_GROUP \
+            else False
+        self.dynamic_host = kwargs.pop('dynamic_host', self.dynamic_host)
+        if self.dynamic_host:
+            self.start_dynamic_host_tracking()
```
armills: We don't want to create a new thread for each cast group. I think what we'd have to do is start the thread for the first group, and add each listening object id to a list. The objects can remove themselves from the list in `__del__`, and when the list is empty the thread can be stopped.

pychromecast/__init__.py
```diff
@@ -334,6 +344,33 @@ def join(self, timeout=None):
...
+        if not name == self.name or not port == self.port:
+            return  # Wrong group, keep discovering
+
+        if host == self.host:
+            return  # No change in host, keep discovering
+
+        self.logger.info('Change in host ip detected, '
+                         'reinitializing object...')
+        stop_discovery(self.browser)
+        self.__init__(host, port=port, device=self.device, dynamic_host=True)
```
armills: We definitely don't want to re-run `__init__` when the host changes. Beyond being bad practice this is going to have a lot of unintended side-effects, such as dropping all the old kwargs passed in.

What we'll want to do is pull everything from init that needs to be re-run when the host changes into a separate method, and call that method in init, and here. ([Probably everything from here down.](https://github.com/balloob/pychromecast/blob/master/pychromecast/__init__.py#L174))

You'll also need to clean-up the old socket client before creating the new one.

pychromecast/__init__.py
freol35241: Thanks for the feedback, the single thread for all groups is a good idea. Had a quick look at it but ran into some issues with the `__del__` -method. It seems like it never gets called. Probably due to using strong (non-weak?) references for the listener registration, see for example [here](https://github.com/balloob/pychromecast/blob/master/pychromecast/__init__.py#L180).

pychromecast/__init__.py
freol35241: Fair enough. Out of interest, is calling the `__init__`-method not recommended due to some python under-the-hood magic that might go bad or is it just to avoid confusion?

What exactly is required for this?
> You'll also need to clean-up the old socket client before creating the new one.

Same as in the [`__del__`-method](https://github.com/balloob/pychromecast/blob/master/pychromecast/__init__.py#L339)?

armills: Yeah, that's pretty much it. AFAIK there isn't a technical reason it can't be called twice, it's just a confusing pattern.

For cleanup, you'll probably just need to call `self.disconnect`. You'll want to keep the default of blocking=True to ensure it closes successfully first.

pychromecast/__init__.py
armills: Yeah, as I understand, CPython doesn't provide any guarantee objects will necessarily be cleaned up in a timely manner. They should at least all be cleaned up with the interpreter is shutting down, though.

balloob:
I think that we should tackle it slightly different. When the group leader changes, it means that the old Chromecast object has been disconnected. We should listen for disconnect events and then try to reconnect, if that fails, try to see if we can find the same Cast Audio somewhere on the network and reconnect there. This should be handled transparently for the consumers of the Cast object.

pychromecast/__init__.py
```diff
@@ -171,8 +178,23 @@ def __init__(self, host, port=None, device=None, **kwargs):
...
+        self.tracker = None
+        self._socket_connection_status = None
+
+        self.setup(host, port, self.device, self.kwargs)
+
+    def setup(self, host, port, device, kwargs):
+        """
+        Method for setting up, and resetting the chromecast object
+        """
+        tries = kwargs.pop('tries', None)
```
freol35241: Another approach would be to default to a certain number (2?) of tries for a cast group and only listen to DISCONNECTED events. Might be a cleaner solution actually. What do you think?

freol35241:
Decided that listening for DISCONNECTED events was a cleaner approach. I am fairly happy with the solution now, and it seems to be stable.

pychromecast/__init__.py
```diff
@@ -122,13 +129,14 @@ class Chromecast(object):
     :param retry_wait: A floating point number specifying how many seconds to
                        wait between each retry. None means to use the default
                        which is 5 seconds.
+    :param dynamic_host: Dynamic or non_dynamic host, if True dynamic host
+                         tracking is enabled. Will be True by default for
+                         cast_type == CAST_TYPE_GROUP.
     """
 
     def __init__(self, host, port=None, device=None, **kwargs):
```
balloob: Instead of storing `kwargs`, please write out all the attributes that we can have and their default values and store as instance variables.

```python
def __init__(self, host, port=None, device=None, *, tries=None, timeout=None, retry_wait=None, blocking=True):
    self._tries = tries
    self._timeout = timeout
```

pychromecast/__init__.py
```diff
@@ -171,8 +179,23 @@ def __init__(self, host, port=None, device=None, **kwargs):
         self.status = None
         self.status_event = threading.Event()
 
+        self.browser = None
+        self.tracker = None
+
+        self.setup(host, port, self.device, self.kwargs)
+
+    def setup(self, host, port, device, kwargs):
```
balloob: Why take in port if that will always be equal to `self.port` ?

balloob: Same with `kwargs` and `device`

pychromecast/__init__.py
```diff
@@ -171,8 +179,23 @@ def __init__(self, host, port=None, device=None, **kwargs):
...
+        """
+        Method for setting up, and resetting the chromecast object
+        """
+        tries = 2 if self.cast_type == CAST_TYPE_GROUP \
```
balloob: Why is tries bigger for a group? We should store these values in the constructor.

Also, please rewrite as normal if statement instead of inline if statement. It is confusing to read.

pychromecast/__init__.py
```diff
@@ -194,6 +217,13 @@ def __init__(self, host, port=None, device=None, **kwargs):
         if blocking:
             self.socket_client.start()
 
+        # Set to True by default if cast_type == CAST_TYPE_GROUP
+        self.dynamic_host = kwargs.pop('dynamic_host', True if
```
balloob: Can be set in the constructor.

pychromecast/__init__.py
```diff
@@ -122,13 +129,14 @@ class Chromecast(object):
     :param retry_wait: A floating point number specifying how many seconds to
                        wait between each retry. None means to use the default
                        which is 5 seconds.
+    :param dynamic_host: Dynamic or non_dynamic host, if True dynamic host
```
balloob: Can you explain what dynamic host tracking is in the comment.

pychromecast/__init__.py
```diff
@@ -334,6 +364,68 @@ def join(self, timeout=None):
...
+            if not self.socket_client.is_connected:
+                # If no host on new ip address was found, try reconnecting
+                self.disconnect()
+                self.setup(self.host, self.port, self.device, self.kwargs)
+
+        self.logger.info('Dynamic host discovery started')
+        self.tracker, self.browser = \
+            start_discovery(self._dynamic_host_callback)
+
+        timeout = threading.Thread(target=discovery_timeout)
```
balloob: We should prevent starting another thread when we already have a thread running.

pychromecast/__init__.py
```diff
@@ -334,6 +364,68 @@ def join(self, timeout=None):
...
+
+            self.logger.info('Discovery timed out, stopping')
+            stop_discovery(self.browser)
+            self.browser = None
+            if not self.socket_client.is_connected:
+                # If no host on new ip address was found, try reconnecting
+                self.disconnect()
+                self.setup(self.host, self.port, self.device, self.kwargs)
+
+        self.logger.info('Dynamic host discovery started')
```
balloob: Can you include the name of the cast device?

pychromecast/__init__.py
```diff
@@ -171,8 +179,23 @@ def __init__(self, host, port=None, device=None, **kwargs):
...
+        """
+        Method for setting up, and resetting the chromecast object
+        """
+        tries = 2 if self.cast_type == CAST_TYPE_GROUP \
+            else kwargs.pop('tries', None)
+        timeout = kwargs.pop('timeout', None)
+        retry_wait = kwargs.pop('retry_wait', None)
+        blocking = kwargs.pop('blocking', True)
+
         self.socket_client = socket_client.SocketClient(
```
balloob: Socket client will automatically try to reconnect. After re-discovery, we should make sure we stop the old socket client or else it will keep reconnecting in the background.

pychromecast/__init__.py
```diff
@@ -334,6 +364,68 @@ def join(self, timeout=None):
...
+
+        self.logger.info('Change in host ip detected, '
+                         'reinitializing object...')
+        stop_discovery(self.browser)
+        self.browser = None
+        self.disconnect()
+        self.host = host
+        try:
+            self.setup(host, port, self.device, self.kwargs)
+        except ChromecastConnectionError:  # noqa
```
balloob: Why `noqa` ?

pychromecast/__init__.py
```diff
@@ -334,6 +364,68 @@ def join(self, timeout=None):
...
+        Callback for socket client connection status
+        """
+        if not status.status == CONNECTION_STATUS_DISCONNECTED:
+            return  # Not interesting
+
+        def discovery_timeout():
+            """
+            Internal scheduled timeout for discovery, 5 seconds
+            """
+            time.sleep(10)
```
balloob: Any reason why we would wait and not just start discovery right away? The function doc also seems to state 5 seconds?

pychromecast/__init__.py
```diff
@@ -334,6 +364,68 @@ def join(self, timeout=None):
...
+            if not self.browser:
+                return  # Change in ip detected, discovery did not time out
+
+            self.logger.info('Discovery timed out, stopping')
```
balloob: You use `logger.info` a lot. I think that for example this one should be reduced to debug level.

pychromecast/__init__.py
balloob: Instead of starting a thread that sleeps for 10 seconds, consider using a [Timer](https://docs.python.org/3.4/library/threading.html#timer-objects)

pychromecast/__init__.py
freol35241: Tries is `None` by default for all other `CAST_TYPES` meaning that the socket client will try to reconnect forever. By defining a finite number of tries we make sure that the socket client will eventually fail (when ip is switched) and then sends a `DISCONNECTED` event which we listen for.

pychromecast/__init__.py
freol35241: As the number of tries is a finite number the old socket client will not try to reconnect forever, also it is stopped [here](https://github.com/balloob/pychromecast/pull/167/commits/d57f10626870df96401e98f579ad35b6653d9405#diff-cd08193135e2ec94e86f3678d2bfdddfR422)

pychromecast/__init__.py
freol35241: Its a timeout to not let the discovery session run forever if no new ip is found.

A timer is a good idea, did not know about that one.

pychromecast/__init__.py
freol35241: Not sure which one you are talking about here. The discovery session or the timeout?

And there will only be one discovery session thread and one timeout thread running simultaneously? Not sure how to otherwise set a timeout on a non-blocking discovery.

pychromecast/__init__.py
freol35241: To be fair, not sure. But flake8 throws an error otherwise.

I guess it has something to do with this [`noqa`](https://github.com/balloob/pychromecast/blob/a867ac9a2cbbdda7f1cd747259009b8dc073e3a2/pychromecast/__init__.py#L13)?

balloob: noqa disabled the linter. What is the error it raises? Why can't it be fixed?

balloob:
Sorry, I have been very busy lately. This PR is still on my radar. Will look at it soon.

In the meanwhile, I just came by a dump of the output of the cast discovery. I wonder that instead of looking at port numbers, we can look at the 'rm' property?

image

To show discovery info on your network, run:

```
pip3 install netdisco==1.0.0rc2
pip3 -m netdisco
```

freol35241:
Sorry, have been busy with other commitments. Will try to find some time soon to finish this PR.

Had a look at the netdisco output on my network and it seems like the `rm` property is false for all the chromecasts? 
image

pychromecast/__init__.py
freol35241: I can be fixed by removing the 
`from .error import *`
in the beginning of the file. Not sure way that import is there though. Do you have any guess?

balloob:
Not sure if that is correct. My Google Home has an `rm`

```
google_cast:
[{'host': '192.168.1.109',
  'hostname': '92f538f5-433e-3500-075f-1f8ae1c92467.local.',
  'port': 8009,
  'properties': {'bs': 'FA8FCA87650A',
                 'ca': '4100',
                 'fn': 'Living room Home',
                 'ic': '/setup/icon.png',
                 'id': '92f538f5433e3500075f1f8ae1c92467',
                 'md': 'Google Home',
                 'rm': '4E05B1EDB9474484',
                 'rs': False,
                 'st': '0',
                 've': '05'}}]
```

foxel:
Hi. Is there any chance this PR will be incorporated some way?

balloob:
Doesn't look like it. Will close it for now as it has gone stale.

foxel:
@balloob is there any plan for implementing this feature? 

I see this lack of cromecast leader tracking as a blocker for home-assistant cast integration to be usable for groups.

PR #180, opened 2017-07-13
Fix send message deadlock

emontnemery:
Correct #179 to avoid throwing exceptions


armills:
Good catch. Thanks for the fix!

PR #184, opened 2017-07-31
Cast web pages with DashCast

madmod:
This is a controller which implements the [DashCast](https://github.com/madmod/dashcast) namespace which allows casting of any web page with an optional refresh interval. (Note the original repo is [here](https://github.com/stestagg/dashcast), but this controller references my deployment of it because I intend to maintain compatibility with this controller.)

~~I had some trouble with the `self.launch` callback not being called in blocking mode, so I hacked around it by waiting for `self.is_active` to be true. This is obviously sub-optimal and if anyone has advice on how to fix it that would be great.~~ Fixed with help from @balloob!

This PR is part of a larger project to allow for customizable Chromecast Backdrops using home-assistant. [Project discussion](https://community.home-assistant.io/t/custom-chromecast-backdrop-overlay-help-welcome/23175)


madmod:
It looks like this fails linting but the log shows the only issues in my PR are the TODOs I've mentioned.

balloob:
This is a great contribution.

I've just fixed all the current linting errors, so there are none in `master` right now. Your PR can't introduce any new ones.

We should find a fix for the callback not getting called instead of trying to hack around it. Can you try if it works with https://github.com/balloob/pychromecast/pull/185 ?

madmod:
Thanks! I'll test it tonight.

.editorconfig
```diff
@@ -0,0 +1,25 @@
+# EditorConfig is awesome: http://EditorConfig.org
```
balloob: Please remove this file.

madmod:
I've decided that [DashCast](https://github.com/madmod/dashcast) is a more polished project with better animations, error message support, and support for loading sites without an iframe.

Currently I'm unable to get error messages or loading without an iframe working, but I think these are issues in the DashCast receiver, not in the controller or pychromecast, so I don't think they should block this PR.

Calling `load_url` twice in a row without using the callback or a delay always fails, but I think that is unusual enough usage that it shouldn't block this PR.

I've added an example/test for the controller in `examples/dashcast_blocking.py`.

I also noticed a typo in `youtube.py` which I fixed, however I'm unable to get it to load a video after launching. (Tested with `examples/blocking.py` before and after my fix.)

I have also added an [EditorConfig](http://editorconfig.org/) file to the repo which encourages text editors to use the current indentation style of four spaces.

I'm unsure if anything should be added to README.rst for this controller. (And for Plex/YouTube also.)

.editorconfig
madmod: I've remove the `.editorconfig`. I'm not sure why you wouldn't want one though. It's a pretty standard part of many open source projects. If you are unfamiliar with it the purpose is to allow anyone to work on the project without having to reconfigure their editor for tabs/spaces when switching between projects.

balloob:
This can be merged once the lint issues have been addressed.

madmod:
I've fixed the linting issues. Because they were in files other than what I wrote (from before my PR) I think someone else should also test them to be sure everything still works well. Thanks!

madmod:
I've fixed the linting issues and I think this is ready to be merged.

pychromecast/controllers/dashcast.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+class DashCastController(BaseController):
+    """ Controller to interact with DashCast app namespace. """
+
+    # pylint: disable=useless-super-delegation
+    # The pylint rule useless-super-delegation doesn't realize
+    # we are setting default values here.
+    def __init__(self,
+                 appNamespace=APP_NAMESPACE,
```
balloob: Why would you set default namespaces and not just hardcode them in the `super` call ?

balloob: (See YouTube controller for example)

madmod: I think it is a good idea to allow others to override the namespace so they can use pychromecast without modifications if they want to deploy their own fork of dashcast. Allowing the app id to be overridden also makes testing a development branch of the receiver easier.

PR #185, opened 2017-08-01
Fix launch callback function

balloob:
Attempt at trying to fix the launch callback function. @madmod, can you verify that this works for your PR?

### DIFF
--- a/pychromecast/socket_client.py
+++ b/pychromecast/socket_client.py
@@ -819,20 +819,23 @@ class ReceiverController(BaseController):
             self.send_message({MESSAGE_TYPE: TYPE_LAUNCH,
                                APP_ID: app_id},
                               callback_function=lambda response:
-                              self._block_till_launched(app_id))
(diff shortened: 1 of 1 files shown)


madmod:
Unfortunately this doesn't work. It does work if the `self.launch` call is not the last thing in the thread. (I added a `time.sleep(10)` after my call to `self.launch`.)

madmod:
I've fixed this in #184 since that seems more efficient than suggesting changes to this PR.

balloob:
What do you mean with last thing in the thread, do you mean last thing in the script?

madmod:
Yeah I meant script. Specifically I tested this by adding those lines to the end of `examples/dashcast_blocking.py`. (Unrelated to the debugging `time.sleep` I currently have in that file for catching any responses from the DashCast receiver. It now works all the time.)

balloob:
Closing this as fix included in #184

PR #194, opened 2017-11-04
Fix link markup in README.rst

erik:
Was originally using markdown syntax.

Surprisingly, rst doesn't allow nested markup, which is why there are two emphasis blocks

http://docutils.sourceforge.net/FAQ.html#is-nested-inline-markup-possible


balloob:
Thanks!

PR #197, opened 2017-11-19
Fix get_device_status for newer cast firmwares

OttoWinter:
See: https://github.com/home-assistant/home-assistant/issues/9965

Newer cast firmwares respond to the `http://CAST_IP:8008/ssdp/device-desc.xml` request pychromecast uses for manual setup with a 404 Not Found error - and so pychromecast _thinks_ the cast device is unavailable. This PR instead uses `http://CAST_IP:8008/setup/eureka_info?options=detail` (found here: http://fiquett.com/2013/07/chromecast-traffic-sniffing/), which should be available on all cast devices/firmwares (tested on Chromecast, Chromecast Audio and Google Home).

The only issue is that I can't seem to find the `api_version` attribute - though it doesn't really seem to be used anywhere, so this shouldn't be a big deal.

### DIFF
--- a/pychromecast/dial.py
+++ b/pychromecast/dial.py
@@ -75,7 +75,8 @@ def get_device_status(host):
         if udn:
             uuid = UUID(udn.replace('-', ''))
 
-        return DeviceStatus(friendly_name, model_name, manufacturer, None, uuid, cast_type)
+        return DeviceStatus(friendly_name, model_name, manufacturer, None,
+                            uuid, cast_type)
 
     except (requests.exceptions.RequestException, ValueError):
         return None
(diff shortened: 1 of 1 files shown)


OttoWinter:
In retrospect, in some cases the `api_version` attribute might be required for some projects - and my current approach doesn't provide that.

So, I have prepared a commit that changes the current behaviour to checking the old way (`/ssdp/device-desc.xml`) first and then falling back to the JSON (`/setup/eureka_info?options=detail`) request. If you think that way is better I would be happy to include it.

balloob:
I'm fine with going with the PR as it currently stands and do a major release to indicate a breaking change. PyChromecast implements the API so a user should not be concerned which API is being used.

balloob:
Thanks! 🐬

balloob:
I tested it with a Google Home and Android TV locally and it worked.

OttoWinter:
Great!

PR #202, opened 2017-12-06
Revert 184 socket client changes

balloob:
This reverts the `socket_client.py` changes introduced in #184 by @madmod because it is breaking the Cast integration in Home Assistant.

### DIFF
--- a/pychromecast/socket_client.py
+++ b/pychromecast/socket_client.py
@@ -819,24 +819,22 @@ class ReceiverController(BaseController):
             self.launch_failure = None
 
             self.send_message({MESSAGE_TYPE: TYPE_LAUNCH,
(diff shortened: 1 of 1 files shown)


madmod:
Its been a while since I implemented this so I’m about fuzzy on the details. What is broken about it? Would it make more sense to update Home Assistant to use the callback?

balloob:
The Home Assistant issue is here: https://github.com/home-assistant/home-assistant/issues/10926

OttoWinter:
Just checked again and this is working with my cast setup + the script I previously used to replicate the bug is now working fine. Thank you!

Can this be merged? I'm desperately waiting for my Home Mini to work in Home Assistant again 😬

PR #203, opened 2017-12-06
Fix CAST_TYPES mapping

### LINKED ISSUE #201: CAST_TYPES dict lacks some cast models
While looking for the culprit for pychromecast breaking with 1.0.2, I have found that some cast devices are not assigned the correct cast type. More precisely, the dictionary in [dial.py](https://github.com/balloob/pychromecast/blob/master/pychromecast/dial.py#L23), that is meant to resolve model names to their corresponding CAST_TYPE [here](https://github.com/balloob/pychromecast/blob/master/pychromecast/dial.py#L71), doesn't include all model name to cast_type mappings. For example, my Google Home Mini reports a `model_name` of "Google Home" and my Chromecast v1 "Eureka Dongle", but none of these keys are in the mapping. This results in pychromecast incorrectly assuming audio-only devices like the Google Home to be the default type (a video+audio chromecast).

This should essentially be an easy fix, since it only involves adding/changing some of the keys in the mentioned mapping - and I would be happy to create a PR. But in order to find the correct mappings, I need some assistance because currently only my Google Home Mini and Chromecast v1 are operational. So if somebody wants to help, just enter http://CAST_IP:8008/setup/eureka_info?options=detail in your browser and report your "model_name" together with the device type (video+audio, or audio-only).

P.S. This issue actually seems to have been around for quite a while. Even the old way of using "/ssdp/device-desc.xml" returned invalid modelNames.

OttoWinter:
Fixes #201 which incorrectly caused google home devices to be assumed to be video+audio cast types by adding new 'google home' and 'eureka dongle' mappings.

### DIFF
--- a/pychromecast/dial.py
+++ b/pychromecast/dial.py
@@ -22,7 +22,9 @@ CAST_TYPE_GROUP = 'group'
 
 CAST_TYPES = {
     u'chromecast': CAST_TYPE_CHROMECAST,
+    u'eureka dongle': CAST_TYPE_CHROMECAST,
     u'chromecast audio': CAST_TYPE_AUDIO,
+    u'google home': CAST_TYPE_AUDIO,
     u'google cast group': CAST_TYPE_GROUP,
 }
 


OttoWinter:
Just noticed that my GitHub client is complaining about README.rst being included in the .gitignore. Is there a reason for that file being ignored? It seems quite odd to me.

balloob:
Yeah that is weird 🤔  you can remove the line in gitignore. I think it's from the time we were auto creating the rst version from the md to satisfy pypi.

balloob:
Thanks for fixing this!

PR #204, opened 2017-12-11
Don't sleep when there are no tries remaining

erik:
Fixes simple logic flow error where the thread would sleep
regardless of there being a remaining retry or not.

### DIFF
--- a/pychromecast/socket_client.py
+++ b/pychromecast/socket_client.py
@@ -276,8 +276,9 @@ class SocketClient(threading.Thread):
                     retry_log_fun = self.logger.debug
 
                     time.sleep(self.retry_wait)
-                    tries -= 1
 
+                if tries:
+                    tries -= 1
         else:
             self.stop.set()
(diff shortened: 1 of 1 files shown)


pychromecast/socket_client.py
```diff
@@ -268,13 +268,16 @@ def initialize_connection(self):
...
                     ConnectionStatus(CONNECTION_STATUS_FAILED,
                                      NetworkAddress(self.host, self.port)))
-                retry_log_fun("Failed to connect, retrying in %.1fs",
-                              self.retry_wait)
-                retry_log_fun = self.logger.debug
 
-                time.sleep(self.retry_wait)
-                if tries:
+                # Only sleep if we have another retry remaining
+                if tries and tries > 1:
```
balloob: If tries is 1, we will now never get to 0.

erik: Hah, duh, sorry.

balloob:
Thanks! 🐬

PR #211, opened 2018-01-19
Drop Python 2 support

balloob:
I have been notified by @jstasiak, the author of ZeroConf, that he is planning to drop support for Python <3.4 in the next release of ZeroConf. So to be ahead of this, here is a PR to drop support for those versions too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
 sudo: false
 language: python
 python:
-  - "2.7"
   - "3.4"
   - "3.5"
+  - "3.6"
 install:
   - pip install -r requirements.txt
   - pip install flake8==3.3.0 pylint==1.7.2
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ python:
   - "3.6"
 install:
   - pip install -r requirements.txt
-  - pip install flake8==3.3.0 pylint==1.7.2
+  - pip install flake8==3.3.0 pylint==1.7.4
 script:
   - flake8 --exclude cast_channel_pb2.py,authority_keys_pb2.py,logging_pb2.py pychromecast
   - pylint pychromecast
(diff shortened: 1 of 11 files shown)


OttoWinter:
The unicode strings in [pychromecast/init.py#341/347](https://github.com/balloob/pychromecast/pull/211/files?diff=split#diff-cd08193135e2ec94e86f3678d2bfdddfL341) can probably be converted to normal strings. Same in pychromecast/socket_client.py#367 and the `CAST_TYPES` dict in pychromecast/dial.py

d8ahazard:
FWIW, I have a fork of zeroconf that reverts the changes to drop V2x support, as well as drop the requirement for netifaces, which fixes this not working on piPy, from what I understand.

While I normally support the need to move onward and upward, in this case, I really think it'd be awesome to continue to support as many versions of python as possible. I for one run it under Plex as a plugin, and they still use v2x in some of their frameworks, which is something I have no control over.

So, I humbly ask you to reconsider dropping 2x support...

balloob:
Sorry @d8ahazard, Python 2 is over. I don't want to depend on a zeroconf fork because there is no guarantee that it will be maintained. Even if you would now say: don't worry, I'll maintain it. Nothing personal but I don't know you. I have good experiences with the original zeroconf being maintained.

OttoWinter:
https://github.com/balloob/pychromecast/blob/4ae1a0c6add9625acb2cc40ff4dd5624ac8a2731/pychromecast/__init__.py#L348-L351
https://github.com/balloob/pychromecast/blob/4ae1a0c6add9625acb2cc40ff4dd5624ac8a2731/pychromecast/__init__.py#L28-L29

can probably be dropped as well.

balloob:
fixed 👍

d8ahazard:
Nooooooooooooooo!  😓

PR #215, opened 2018-02-14
Fixes for discovery, FreeBSD Socket creation

d8ahazard:
Fix issues in discovery where things blow up if browser is not able to be created because of networking issues.

Add proper checks in socket_client for BSD-derived operating systems, bind sockets accordingly...

Lint

### DIFF
--- a/pychromecast/discovery.py
+++ b/pychromecast/discovery.py
@@ -9,6 +9,7 @@ DISCOVER_TIMEOUT = 5
 
 class CastListener(object):
     """Zeroconf Cast Services collection."""
+
     def __init__(self, callback=None):
         self.services = {}
         self.callback = callback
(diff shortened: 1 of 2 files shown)


pychromecast/discovery.py
```diff
@@ -84,8 +85,13 @@ def start_discovery(callback=None):
...
     listener = CastListener(callback)
-    return listener, \
-        ServiceBrowser(Zeroconf(), "_googlecast._tcp.local.", listener)
+    sb = False
+    try:
+        sb = ServiceBrowser(Zeroconf(), "_googlecast._tcp.local.", listener)
+    except Exception:
+        return listener, False
+    finally:
+        return listener, sb
```
OttoWinter: ⚠️because of this statement in `finally`, `return listener, False` will never be returned.

carbon 3

I mean it doesn't really affect the return value, it's still misleading though.

pychromecast/discovery.py
```diff
@@ -110,5 +117,8 @@ def callback(name):
         discover_complete.wait(timeout)
 
         return listener.devices
+    except Exception:
+        return
     finally:
-        stop_discovery(browser)
+        if browser is not False:
```
OttoWinter: Isn't browser *always* `False`? 👀

d8ahazard:
Updated return to accomplish what I wanted, add 'proper' exception types...

>>> def f():
...     foo = 'foo'
...     try:
...             raise ValueError()
...             print('try')
...     except:
...             print('except')
...             foo = False
...     finally:
...             print('finally')
...             return foo
...
>>> f()
except
finally
False
>>>

balloob:
ok to merge when tests passes.

PR #221, opened 2018-03-17
Add adjusted_current_time() function

tnyeanderson:
This pull request adds a timestamp (`last_updated`) to the `MediaStatus` class. This attribute is used in the method `MediaStatus.adjusted_current_time()` which returns the approximate current seek time.

It is unknown whether buffering would cause this to return an inaccurate seek time. I am not able to test this because I am fortunate enough to have fast internet! From what it seems, however, since the `MediaStatus` would be updated each time buffering or playing occurs, it should work just fine.

### DIFF
--- a/pychromecast/controllers/media.py
+++ b/pychromecast/controllers/media.py
@@ -76,7 +76,7 @@ class MediaStatus(object):
         if self.player_state == MEDIA_PLAYER_STATE_PLAYING:
             # Add time since last update
             return (self.current_time +
-                    datetime.utcnow() - self.last_updated).total_seconds()
+                    (datetime.utcnow() - self.last_updated)).total_seconds()
         # Not playing, return last reported seek time
         return self.current_time
 
--- a/pychromecast/controllers/media.py
+++ b/pychromecast/controllers/media.py
@@ -71,8 +71,9 @@ class MediaStatus(object):
         self.current_subtitle_tracks = []
         self.last_updated = None
 
-    def get_seek_time(self):
-        """ Returns current seek time of media in seconds """
+    @property
+    def adjusted_current_time(self):
+        """ Returns calculated current seek time of media in seconds """
         if self.player_state == MEDIA_PLAYER_STATE_PLAYING:
             # Add time since last update
             return (self.current_time +
--- a/pychromecast/controllers/media.py
+++ b/pychromecast/controllers/media.py
@@ -76,7 +76,7 @@ class MediaStatus(object):
         if self.player_state == MEDIA_PLAYER_STATE_PLAYING:
             # Add time since last update
             return (self.current_time +
-                (datetime.utcnow()-self.last_updated).total_seconds())
+                    datetime.utcnow() - self.last_updated).total_seconds()
         # Not playing, return last reported seek time
         return self.current_time
 
--- a/pychromecast/controllers/media.py
+++ b/pychromecast/controllers/media.py
@@ -73,8 +73,7 @@ class MediaStatus(object):
 
     def get_current_time(self):
         """ Get current seek time of video (does not account for buffering time) """
-        if :
-            return (self.current_time + (time.time()-self.last_updated)) if self.player_state == MEDIA_PLAYER_STATE_PLAYING else self.current_time
+        return (self.current_time + (time.time()-self.last_updated)) if self.player_state == MEDIA_PLAYER_STATE_PLAYING else self.current_time
 
     @property
     def metadata_type(self):
--- a/pychromecast/controllers/media.py
+++ b/pychromecast/controllers/media.py
@@ -72,7 +72,7 @@ class MediaStatus(object):
         self.last_updated = time.time()
 
     def get_current_time(self):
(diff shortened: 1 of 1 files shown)


pychromecast/controllers/media.py
```diff
@@ -67,6 +69,15 @@ def __init__(self):
         self.media_metadata = {}
         self.subtitle_tracks = {}
         self.current_subtitle_tracks = []
+        self.last_updated = time.time()
```
OttoWinter: I think we should have `last_updated` be invalid (i.e. `None`) when no information has been received yet to indicate **unknown state**.

pychromecast/controllers/media.py
```diff
@@ -67,6 +69,15 @@ def __init__(self):
...
+
+    def get_current_time(self):
+        """ Get current seek time of video """
```
OttoWinter: 1. It's not always a "video", it can also be audio; better change it to `Get current seek time of media`.
2. The method name is not very nice - `get_current_time` doesn't suggest seek times. Maybe `get_seek_time`.
3. Please provide a better docstring; i know this library is not too great with documentation, but still it would be good to specify what this method returns - the time *in seconds* since start of media/seek time.
4. What should this method return if nothing is playing? I think we should return `None` or something like that to indicate invalid call then.

pychromecast/controllers/media.py
```diff
@@ -67,6 +69,15 @@ def __init__(self):
...
+        if self.player_state == MEDIA_PLAYER_STATE_PLAYING:
+            current_time = self.current_time + (time.time()-self.last_updated)
+        else:
+            current_time = self.current_time
+        return current_time
```
OttoWinter: I personally would have made this `return` statement inline, it's a bit more readable:

```python
if self.player_state == MEDIA_PLAYER_STATE_PLAYING:
    # Add time since last update
    return self.current_time + (time.time()-self.last_updated)
# Not playing, return last reported seek time
return self.current_time
```

pychromecast/controllers/media.py
```diff
@@ -232,6 +243,7 @@ def update(self, data):
         self.subtitle_tracks = media_data.get('tracks', self.subtitle_tracks)
         self.current_subtitle_tracks = status_data.get(
             'activeTrackIds', self.current_subtitle_tracks)
+        self.last_updated = time.time()
```
OttoWinter: I'm not entirely sold on the use of the `time` module here. I personally like more descriptive types than just floats/integers passed around. Maybe have a look at [`datetime.utcnow()`](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow) + returning [`datetime.timedelta`](https://docs.python.org/3/library/datetime.html#timedelta-objects) as return value?

tnyeanderson:
That should do it. Let me know if you think there's anything else it might need. Love the program, thanks for putting in all the work!

pychromecast/controllers/media.py
```diff
@@ -67,6 +69,16 @@ def __init__(self):
...
         self.subtitle_tracks = {}
         self.current_subtitle_tracks = []
+        self.last_updated = None
+
+    def get_seek_time(self):
+        """ Returns current seek time of media in seconds """
+        if self.player_state == MEDIA_PLAYER_STATE_PLAYING:
+            # Add time since last update
+            return (self.current_time +
+                    datetime.utcnow() - self.last_updated).total_seconds()
```
balloob: I'm a bit surprised here, shouldn't you put parentheses around the `now - last_updated` part?

tnyeanderson: Technically it doesn't matter, and the line break provides some clarity... I did take them out because it made the line break indentation look weird. Want them back in?

tnyeanderson: This is my first time really using python (I'm usually a JS guy) so sorry about any bad practices I may be making!

tnyeanderson: (First pull request too actually... hopefully many more ahead!)

balloob: let's keep them around it, code is more often read than that it is written.

balloob: Welcome to open source, you're doing great work 🎉

pychromecast/controllers/media.py
```diff
@@ -67,6 +69,16 @@ def __init__(self):
         self.media_metadata = {}
         self.subtitle_tracks = {}
         self.current_subtitle_tracks = []
+        self.last_updated = None
+
+    def get_seek_time(self):
```
balloob: I am not sure about the name 'seek_time'. Also, why would we make it a method and not just a property like the other ones.

I would expect it to be:
 - called `adjusted_current_time`. Make sure to add in the comment why it is called adjusted (it calculates current time)
 - be a property

tnyeanderson: I had it as a property originally, but it wasn't working (likely just beginner error, still figuring things out). I''ll look into the differences between properties and methods in python. Thanks for all the advice!

PR #222, opened 2018-03-18
Fix not waiting with no tries set

OttoWinter:
We're not waiting when `tries` is `None` but `retry_wait` is set. This can cause a flood of requests if no connection can be made.

Probably related https://github.com/home-assistant/home-assistant/issues/13319

(I've been using this patch for a while now and apparently forgot to put it in a PR)

### DIFF
--- a/pychromecast/socket_client.py
+++ b/pychromecast/socket_client.py
@@ -267,7 +267,7 @@ class SocketClient(threading.Thread):
                                      NetworkAddress(self.host, self.port)))
 
                 # Only sleep if we have another retry remaining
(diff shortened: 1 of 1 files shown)


pychromecast/socket_client.py
```diff
@@ -267,11 +267,10 @@ def initialize_connection(self):
                                      NetworkAddress(self.host, self.port)))
 
                 # Only sleep if we have another retry remaining
-                if tries and tries > 1:
+                if (tries and tries > 1) or not tries:
```
balloob:
```python
if tries is None or tries > 1:
```

PR #224, opened 2018-03-20
Fixed misplaced parenthesis

tnyeanderson:
Misplaced parenthesis in adjusted_current_time() was causing a TypeError.

### DIFF
--- a/pychromecast/controllers/media.py
+++ b/pychromecast/controllers/media.py
@@ -77,7 +77,7 @@ class MediaStatus(object):
         if self.player_state == MEDIA_PLAYER_STATE_PLAYING:
             # Add time since last update
             return (self.current_time +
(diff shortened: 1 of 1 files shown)


tnyeanderson:
Thanks for being patient while I figured this out! You got me wanting to do more open source work soon.

balloob:
Every time you open a new PR, make sure you base your work off the latest main branch (in this case `master`). That way you also know that your work works with the latest version.

balloob:
This bugfix shows that we could use some tests for pychromecast ;)

tnyeanderson:
Got it, I'll make sure to do that in the future! Hard to make a test script because everyone's chromecast is different (and it would probably take a long time to run the tests)... But would definitely be useful. Sorry for spoiling a release with buggy code!

PR #227, opened 2018-03-27
Add Spotify Controller

enriquegh:
One thing to note is that this uses Spotipy which at the moment is not updated on PyPi. See https://github.com/plamere/spotipy/issues/211.

### DIFF
--- a/pychromecast/controllers/spotify.py
+++ b/pychromecast/controllers/spotify.py
@@ -1,22 +1,96 @@
-from pychromecast.controllers import BaseController
 import logging
+import os
+import requests
+
+from pychromecast.controllers import BaseController
+from ..config import APP_SPOTIFY
 
 logging.basicConfig(level=logging.DEBUG)
 
+APP_NAMESPACE = "urn:x-cast:com.spotify.chromecast.secure.v1"
+USER_AGENT = "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_2) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/63.0.3239.132 Safari/537.36"
+
+TYPE_STATUS = "setCredentials"
+
 
 class SpotifyController(BaseController):
-    def __init__(self):
+    def __init__(self, username, password, friendly_name):
         super(SpotifyController, self).__init__(
-            "urn:x-cast:com.spotify.chromecast.secure.v1", "CC32E753")
+              APP_NAMESPACE, APP_SPOTIFY)
+
+        self.logger = logging.getLogger(__name__)
+        self.username = username
+        self.password = password
+        self.session_started = False
+        self.token = None
+        self.expiration_date = None
 
     def receive_message(self, message, data):
-        print("Cast message: {}".format(data))
+        self.logger.debug("Cast message: {}".format(data))
+        
+        return True
 
-        return True  # indicate you handled this message
+    def _get_csrf(self, session, cookies):
+        headers = {'user-agent': USER_AGENT}
 
-    def start_playback(self, credentials):
+        response = session.get("https://accounts.spotify.com/login",
+                               headers=headers, cookies=cookies)
+        response.raise_for_status()
+
+        return response.cookies['csrf_token']
+
+    def _login(self, session, cookies, username, password, csrf_token):
+        headers = {'user-agent': USER_AGENT}
+
+        data = {"remember": False, "username": username, "password": password,
+                "csrf_token": csrf_token}
+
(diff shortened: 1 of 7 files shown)


balloob:
Nice!

I think however that most of this code should not be in PyChromecast. I think that the Spotify Controller should be constructed with just the auth token. Getting a valid authentication token is out of scope for PyChromecast. That will also make sure that we don't have to include a new dependency for a controller.

There is also a bunch of commits in this branch unrelated to this feature, they need to be removed

pychromecast/controllers/spotify.py
```diff
@@ -0,0 +1,169 @@
...
+            self.logger.debug("Using same token: %s", self.token)
+            self.launch(callback_function=callback)
+        else:
+            self.logger.debug("Creating new token")
+            self.start_session()
+            self.logger.debug("Token is: %s", self.token)
+            self.launch(callback_function=callback)
+
+        # Need to wait for Spotify to be launched on Chromecast completely
+        time.sleep(5)
```
OttoWinter: Can't we poll here, maybe even using `receive_message` or your callback defined above? Using `time.sleep` seems like quite the hack to me...

pychromecast/controllers/spotify.py
```diff
@@ -0,0 +1,169 @@
...
+            if device['name'] == self.cast_name:
+                return device['id']
+        return None
+
+    def play_song(self, uri):
+        """ Play a single song with it's Spotify URI. """
+        if self.device_id is None:
+            raise Exception("No device id. Try launching app again")
+        else:
+            self.client.start_playback(device_id=self.device_id, uris=[uri])
```
OttoWinter: Personally, I would unindent this code block as above expression already `raise`s. See http://blog.timoxley.com/post/47041269194/avoid-else-return-early

pychromecast/controllers/spotify.py
```diff
@@ -0,0 +1,169 @@
...
+            self.logger.debug("Token is: %s", self.token)
+            self.launch(callback_function=callback)
+
+        # Need to wait for Spotify to be launched on Chromecast completely
+        time.sleep(5)
+        self.client = spotipy.Spotify(auth=self.token)
+
+        self.device_id = self.get_spotify_device_id()
+
+    def get_spotify_device_id(self):
```
OttoWinter: I took a look at supporting Spotify some time ago, and if I remember correctly spotify sends the device Id under the  `urn:x-cast:com.spotify.chromecast...` namespace.

pychromecast/controllers/spotify.py
```diff
@@ -0,0 +1,169 @@
...
+        """
+        Gets device id from Spotify. This should be the same as the name of
+        the Chromecast.
+        """
+        devices_available = self.client.devices()
+
+        for device in devices_available['devices']:
+            self.logger.debug(device)
+            if device['name'] == self.cast_name:
```
OttoWinter: Does `spotipy`'s `devices()` method return something like `UUID`? Because (almost) all chromecasts have a UUID and I think this code would be more robust if we wouldn't solely check the name.

enriquegh: There is an id field which is the one that is being used to send the music to but afaik these can change at any moment.
The only things tying the device on the Spotify API and the Chromecast is name and perhaps the type which is set to `CastVideo`. Would checking for type to be equal to `CastVideo` make this a bit better?

pychromecast/controllers/spotify.py
enriquegh: Looking at the messages sent to the `urn:x-cast:com.spotify.chromecast.secure.v1` namespace, there's a `session_id` and a `transport_id` but they don't match the device id we need.

pychromecast/controllers/spotify.py
enriquegh: That might be much better. Any pointers on how to poll this? Not sure how to poll the socket_client for when the app is fully initialized?

enriquegh:
@balloob what would you propose? Have a complete separate repo that gets the auth token and only maybe point to it?
The unfortunate thing is that none of the auth tokens that you can request via Spotify's Web API (hence through spotipy as well) have enough permissions to start the session on the chromecast.

Regarding the commits, I pulled from upstream and a mess was created. Not sure how to delete them? Would you happen to know?

balloob:
Easiest to clean up the mess would be to create a new branch based off `master` and cherry pick your commits.

I would indeed propose a separate repo for retrieving the auth token. That code should not belong in PyChromecast.

PR #230, opened 2018-04-08
Spotify controller creation

enriquegh:
Continuation of #227 
Previous commits have been removed and only Spotify Controller commits remain.

I implemented some of the suggestions made on the other pull request like:
- Poll the received messages for `setCredentialsResponse`. This can help be more accurate for when the Spotify up is completely finished.
- Removed the access token logic to it's own repository

Let me know if you have any questions or suggestions.
Thanks!

### DIFF
--- a/examples/spotify_example.py
+++ b/examples/spotify_example.py
@@ -1,0 +1,14 @@
+import pychromecast
+from pychromecast.controllers.spotify import SpotifyController
+
+chromecasts = pychromecast.get_chromecasts()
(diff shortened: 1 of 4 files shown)


balloob:
Yeah, adding it to the example makes sense. Just mention in the example description where to install dependencies from, so we don't have to add it as dependencies of pychromecast.

enriquegh:
The example uses spotipy and spotify-token but they are not added to requirements.txt.

PR #236, opened 2018-05-20
Better YouTube controller

ur1katz:
Adds the following functionality to the YouTube controller:
* Play video
* Add video to video queue
* Play next - play a video after the current video ends
* Remove video  from queue


balloob:
@OttoWinter I don't currently have access to a Chromecast + YouTube. Would you have time to do a quick test to see if this works?

pychromecast/controllers/youtube.py
```diff
@@ -1,59 +1,252 @@
...
 from . import BaseController
+from ..error import UnsupportedNamespace
 
-MESSAGE_TYPE = "type"
+YOUTUBE_BASE_URL = "https://www.youtube.com/"
+BIND_URL = YOUTUBE_BASE_URL + "api/lounge/bc/bind"
+LOUNGE_TOKEN_URL = YOUTUBE_BASE_URL +\
+                   "api/lounge/pairing/get_lounge_token_batch"
+
+HEADERS = {"Origin": YOUTUBE_BASE_URL,
```
balloob: Please extract all the logic to communicate directly with the YouTube servers into a standalone Python package that will be installed from PyPI

balloob:
Yeah, if you can help testing that would be great. 

I do think that we should extract all the HTTP YouTube stuff into a standalone package. We can still use it inside PyChromecast, but can make it an extra install. I don't want to get PRs for every change YouTube makes, rather just have to bump a version number.

theychx:
Everything appears to be working reliably.
(```play_video```, ```add_to_queue```, ```play_next```, ```remove_video```)

ur1katz:
I moved the logic to the casttube package.

I couldn't find a good way of hooking to the register_controller to start the session so i ended up calling "start_session_if_none"  function before each session call. That came out  pretty ugly so i would appreciate any other suggestions..

@theychx Thanks for the extra testing!

balloob:
Awesome!

PR #238, opened 2018-06-01
Create spotify_exemple_2.py

SamuelPoggioli:
Script for test pychromecast with Spotify but with the manual entry of the token. Because the dependency spotify_token method don't work ( Spotify added a reCAPTCHA button).  To test because SpotifyController don't work for me...


balloob:
If the old one no longer works, we should replace it instead of keeping it around. 

CC @enriquegh

balloob:
@enriquegh now you can see why I did not want that code in pychromecast. Too fragile

balloob:
No worries. I've been burned before with the development of @home-assistant.

PR #239, opened 2018-06-04
Spotify authorization Code Flow

SamuelPoggioli:
Remove spotify-token usage and add Spotify authorization Code Flow
This token acquisition method is less easy to use but it is more robust.

### DIFF
--- a/examples/spotify_example.py
+++ b/examples/spotify_example.py
@@ -3,12 +3,19 @@ Example on how to use the Spotify Controller.
 NOTE: You need to install the spotipy and spotify-token dependencies.
 
 This can be done by running the following:
-pip install spotify-token
 pip install git+https://github.com/plamere/spotipy.git
+
+You must declare your application to Spotify
+https://developer.spotify.com/dashboard/
+When you have create your app, you must specify the "Redirect URIs"
+in "EDIT SETTINGS".
+Ex value : http://localhost
+At the first launch you will be redirected to your browser to 
+allow the application.
 """
 import pychromecast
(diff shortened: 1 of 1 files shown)


balloob:
@enriquegh is this ok?

enriquegh:
last time I tried getting a token from spotipy this didn't work.
It seems that the access token doesn't have enough permissions to start the Spotify session on Chromecast.

Did this work for you @SamuelPoggioli ?
Let me give this a shot, maybe something changed there since I last did this.

SamuelPoggioli:
If you use the "Client Credentials Flow" method, the token does not have the necessary privileges but I use the "Authorization Code Flow" method which has all the necessary privileges (the "SCOPE" variable defines the desired privileges).

The problem with this method is that on first use, the script must use the web browser to request a usage token (a other...). Spotify asks to validate the application and returns the usage token in the URL's GET parameters. 

The credentials are cached locally and are used to automatically re-authorized expired tokens.

Nojahhh:
Any progress regarding this issue?

emontnemery:
@fondberg I don't think this is needed, #275 should be enough?

Nojahhh:
Can confirm. #275 solved it for me. 
Thanks a lot @fondberg and @emontnemery.

fondberg:
I don't think this is needed in the short term and I believe it has a flaw in that it requires a user confirm which is not so easy in a backend solution.

For the future when spotify adds more oauth realms we can switch to that. In the meantime I believe impersonating a browser which spotipy does is ok

emontnemery:
Closing in favour of #275.

Issue #243, opened 2018-06-27
socket loop breaks when application launch raise exception (via callback)


guirem:
When exception occurs in callback function, it breaks the socket_client run() loop.
I got the case when using media.block_until_active() when running a new application. Exception is raised by _block_till_launched function of socket_client

Error is coming from the cast device for whatever reason (application specific?) so should not break the  socket loop but should be managed via the existing launch error listener (register_launch_error_listener)

```
# line 398 - socket_client.py
# existing code
if REQUEST_ID in data:
            callback = self._request_callbacks.pop(data[REQUEST_ID], None)
            if callback is not None:
                event = callback['event']
                callback['response'] = data
                function = callback['function']
                event.set()
                if function:
                    function(data)

# proposition
if REQUEST_ID in data:
            callback = self._request_callbacks.pop(data[REQUEST_ID], None)
            if callback is not None:
                event = callback['event']
                callback['response'] = data
                function = callback['function']
                event.set()
                if function:
                    try:
                        function(data)
                    except Exception as e :
                        self.logger.error("runonce: Error calling callback  : %s" % str(e))
                        # just log and do nothing else to keep the loop running....

```

guirem:
Pull request with proposal : [pull request #244 ](https://github.com/balloob/pychromecast/pull/244)

PR #244, opened 2018-06-27
fix for issue #243

guirem:
Extend simple_listener_example.py with more listeners + fix proposal for exception breaking the socket loop.

### DIFF
--- a/pychromecast/socket_client.py
+++ b/pychromecast/socket_client.py
@@ -405,7 +405,7 @@ class SocketClient(threading.Thread):
                 if function:
                     try:
                         function(data)
-                    except Exception as e :
+                    except Exception as e:
                         self.logger.error("runonce: Error calling callback  : %s" % str(e))
                         # maybe do something else...
 
--- a/pychromecast/socket_client.py
+++ b/pychromecast/socket_client.py
@@ -403,7 +403,11 @@ class SocketClient(threading.Thread):
                 function = callback['function']
                 event.set()
                 if function:
-                    function(data)
+                    try:
+                        function(data)
+                    except Exception as e :
+                        self.logger.error("runonce: Error calling callback  : %s" % str(e))
+                        # maybe do something else...
 
         return 0
 
--- a/examples/simple_listener_example.py
+++ b/examples/simple_listener_example.py
@@ -25,6 +25,28 @@ class StatusMediaListener:
     def new_media_status(self, status):
         print('[',time.ctime(),' - ', self.name,'] status media change:')
         print(status)
+        
+class ConnectionListener:
+    def __init__(self, name, cast):
+        self.name = name
+        self.cast= cast
+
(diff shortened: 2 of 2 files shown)


pychromecast/socket_client.py
```diff
@@ -403,7 +403,11 @@ def run_once(self):
                 function = callback['function']
                 event.set()
                 if function:
```
balloob: Could you rewrite this code block to use guard clauses and thus return early?

```python
if REQUEST_ID not in data:
    return 0

callback = …

if callback is None:
    return 0

function = …

if not function:
    return 0

try:
    function(data)
except Exception:
    self.logger.exception('runonce: error calling callback')

return 0
```

pychromecast/socket_client.py
```diff
@@ -403,7 +403,11 @@ def run_once(self):
...
-                    function(data)
+                    try:
+                        function(data)
+                    except Exception as e:
+                        self.logger.error("runonce: Error calling callback  : %s" % str(e))
+                        # maybe do something else...
```
balloob: Should we close the connection and return 1?

guirem: It actually depends on for what kind of action the callback is required. It depends of the code in the callback function.

In my case it was when launching an app and error is on google cast device side so it should not close the connection (so return 0). It happened on some case when an application is locked on the cast or if app id is not found.
The 'new_launch_error' listener should implement the logic to handle the correct behavior when it happens but this is not a socket issue.

Maybe in some other cases, the cause is more serious and could require disconnection but I'm not really an expert in this.
My guess would be to never disconnect.

I also believe that this could explained some previous cases I experienced when the callback failed (for whatever reason) and was causing loss of control with pychromecast. 
It could be judicious to show in the logs the underlying/parent exception for debug purpose. The `self.logger.exception('runonce: error calling callback')` will not give enough information.

An alternative way would be remove the exception raised on this part as it's not really relevant in that case (it's just a regular error status and not an exception).
```
# original
def _block_till_launched(self, app_id):
        if self.blocking:
            self.app_launch_event.wait()
            if self.launch_failure:
                raise LaunchError(
                    "Failed to launch app: {}, Reason: {}".format(
                        app_id, self.launch_failure.reason))

# just logging instead of raising exception
def _block_till_launched(self, app_id):
        if self.blocking:
            self.app_launch_event.wait()
            if self.launch_failure:
                self.logger.error(
                    "Failed to launch app: {}, Reason: {}".format(
                        app_id, self.launch_failure.reason))
```

It's kind of up to you.

balloob: `self.logger.exception` prints the full stack trace of all exceptions, it will give all the needed information.

I like failing hard. So if the app that is using Chromecast is not handling something correctly, we should blow up . That way the developer knows their app is not working as expected.

PR #265, opened 2019-02-07
Don't throw when closing socket.

emontnemery:
Attempt to improve stability of audio groups (home-assistant/home-assistant#13530)

Root cause of home-assistant/home-assistant#13530 is that Home Assistant does not reconnect after group leader change. The issue is easy to reproduce by rebooting members of an audio group repeatedly.

Several times, I've seen it happening after failing to cleanup the socket client:
```
Exception in thread Thread-30:
  File "/usr/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/home/erik/github/home-assistant_fork/lib/python3.6/site-packages/pychromecast/socket_client.py", line 348, in run
    self._cleanup()
  File "/home/erik/github/home-assistant_fork/lib/python3.6/site-packages/pychromecast/socket_client.py", line 496, in _cleanup
    self.socket.close()
AttributeError: 'NoneType' object has no attribute 'close'
```

With this change the stability seems to be significantly improved, although logs indicate there might also be bugs in python-zeroconf.

Edit: Not a fix, still possible to trigger the bug.

### DIFF
--- a/pychromecast/socket_client.py
+++ b/pychromecast/socket_client.py
@@ -6,7 +6,7 @@ version of this code: https://github.com/minektur/chromecast-python-poc.
 Without him this would not have been possible.
 """
 # Pylint does not understand the protobuf objects correctly
-# pylint: disable=no-member
+# pylint: disable=no-member, C0302
 
 import errno
 import json
--- a/pychromecast/socket_client.py
+++ b/pychromecast/socket_client.py
@@ -490,7 +490,11 @@ class SocketClient(threading.Thread):
             except Exception:  # pylint: disable=broad-except
                 pass
 
-        self.socket.close()
+        try:
+            self.socket.close()
+        except Exception:  # pylint: disable=broad-except
+            pass
+
         self._report_connection_status(
             ConnectionStatus(CONNECTION_STATUS_DISCONNECTED,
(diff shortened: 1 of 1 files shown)


pychromecast/socket_client.py
```diff
@@ -6,7 +6,7 @@
 Without him this would not have been possible.
 """
 # Pylint does not understand the protobuf objects correctly
-# pylint: disable=no-member
+# pylint: disable=no-member, C0302
```
OttoWinter: What is `C0302`? Can you use the full name?

emontnemery:
@OttoWinter No it's not a complete fix, I'm working on one, but it's a bit more work. 

In short, the service browser we use from pychromecast will update when a service is added or removed. The service only includes the service name, not host name, IP, or port. Hence pychromecast and the Hass integration has to be rewritten a bit with that in mind (we should not resolve hostname and port when we get the service update but just before connecting).

I have something working, but need some time to test and cleanup the code.

masterz666:
Hi, 

I'm affected by the issue and if you need someone to perform some tests, I can help with that.
just msg me with some guidance on how to integrate you code in my HA test instance. 

Thx.

emontnemery:
I'm closing this in favor of #267

PR #267, opened 2019-02-10
Add support for connecting to named service

emontnemery:
Add support for connecting to a named service.

This is helpful for audio groups where host name may change over time.
Also:
- Don't throw when closing socket
- Improve prints

### DIFF
--- a/pychromecast/dial.py
+++ b/pychromecast/dial.py
@@ -7,7 +7,7 @@ from uuid import UUID
 import logging
 import requests
 
-from .discovery import get_host_from_service
+from .discovery import get_info_from_service
 
 XML_NS_UPNP_DEVICE = "{urn:schemas-upnp-org:device-1-0}"
 
@@ -51,7 +51,9 @@ def get_device_status(host, services=None, zconf=None):
     try:
         if not host:
             for service in services.copy():
-                host, _ = get_host_from_service(service, zconf)
+                service_info = get_info_from_service(service, zconf)
+                if service_info and service_info.server:
+                    host = service_info.server.lower()
                 if host:
                     _LOGGER.debug("Resolved service %s to %s", service, host)
                     break
--- a/pychromecast/dial.py
+++ b/pychromecast/dial.py
@@ -33,6 +33,7 @@ CAST_TYPES = {
 
 _LOGGER = logging.getLogger(__name__)
 
+
 def reboot(host):
     """ Reboots the chromecast. """
     CC_SESSION.post(FORMAT_BASE_URL.format(host) + "/setup/reboot",
--- a/pychromecast/dial.py
+++ b/pychromecast/dial.py
@@ -4,7 +4,10 @@ Implements the DIAL-protocol to communicate with the Chromecast
 from collections import namedtuple
 from uuid import UUID
 
+import logging
 import requests
+
+from .discovery import get_host_from_service
 
 XML_NS_UPNP_DEVICE = "{urn:schemas-upnp-org:device-1-0}"
 
@@ -28,6 +31,7 @@ CAST_TYPES = {
     'google cast group': CAST_TYPE_GROUP,
 }
 
+_LOGGER = logging.getLogger(__name__)
 
 def reboot(host):
     """ Reboots the chromecast. """
@@ -35,7 +39,7 @@ def reboot(host):
                     data='{"params":"now"}', timeout=10)
 
 
-def get_device_status(host):
+def get_device_status(host, services=None, zconf=None):
     """
     :param host: Hostname or ip to fetch status from
     :type host: str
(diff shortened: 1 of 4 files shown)


pychromecast/socket_client.py
```diff
@@ -235,50 +255,65 @@ def initialize_connection(self):
...
-                    retry_log_fun = self.logger.debug
-                    time.sleep(self.retry_wait)
-
-                if tries:
-                    tries -= 1
-        else:
-            self.stop.set()
-            self.logger.error("Failed to connect. No retries.")
-            raise ChromecastConnectionError("Failed to connect")
+            for service in self.services.copy():
```
emontnemery: Loop over an iterable with services, take care to copy since items may be added or removed.

pychromecast/socket_client.py
```diff
@@ -212,6 +217,21 @@ def __init__(self, host, port=None, cast_type=CAST_TYPE_CHROMECAST,
                                  NetworkAddress(self.host, self.port)))
             raise
 
+    def _get_host_from_service(self, service):
+        """ Resolve host and port from service. """
+        host = self.host
+        port = self.port
+        if service:
```
emontnemery: if service is `None`, we're connecting directly to a host name or IP-address

pychromecast/socket_client.py
```diff
@@ -235,50 +255,65 @@ def initialize_connection(self):
...
+                            "aborting due to stop signal.",
+                            self.host, self.port, err)
+                        raise ChromecastConnectionError("Failed to connect")
+
+                    self._report_connection_status(
+                        ConnectionStatus(CONNECTION_STATUS_FAILED,
+                                         NetworkAddress(self.host, self.port)))
+
+            # Only sleep if we have another retry remaining
+            if tries is None or tries > 1:
```
emontnemery: Sleep in the while loop to handle an empty `self.services`

pychromecast/dial.py
```diff
@@ -44,6 +49,13 @@ def get_device_status(host):
     """
 
     try:
+        if not host:
+            for service in services.copy():
```
balloob: This works. I usually prefer `list(services)` if I'm afraid the source will be mutated during iteration.

pychromecast/socket_client.py
balloob: Can you add this as a comment in the code?

pychromecast/socket_client.py
```diff
@@ -212,6 +217,21 @@ def __init__(self, host, port=None, cast_type=CAST_TYPE_CHROMECAST,
...
+        host = self.host
+        port = self.port
+        if service:
+            host, port = get_host_from_service(service, self.zconf)
+            if host and port:
+                self.logger.debug("[%s:%s] Resolved service %s to %s:%s",
+                                  self.host, self.port, service, self.host,
+                                  self.port)
+            else:
+                self.logger.debug("[%s:%s] failed to resolve service %s",
```
balloob: Should we do something besides debug logging that we failed to resolve?

pychromecast/discovery.py
```diff
@@ -128,3 +137,19 @@ def callback(name):
...
+    """ Resolve host and port from service. """
+    host_info = (None, None)
+    try:
+        service_info = zconf.get_service_info('_googlecast._tcp.local.',
+                                              service)
+        if service_info and service_info.server and service_info.port:
+            _LOGGER.debug(
+                "get_host_from_service resolved service %s to service_info %s",
+                service, service_info)
+            host_info = (service_info.server.lower(), service_info.port)
```
balloob: Any reason we wouldn't just return `service_info`  ?

balloob: And then just return either `None` or raise when fetching fails.

pychromecast/socket_client.py
```diff
@@ -212,6 +217,21 @@ def __init__(self, host, port=None, cast_type=CAST_TYPE_CHROMECAST,
                                  NetworkAddress(self.host, self.port)))
             raise
 
+    def _get_host_from_service(self, service):
```
balloob: This wrapper feels weird. It adds a code indirection that is a bit confusing. Can't we have the caller of this method directly call `get_host_from_service` ? 

I also think that in the `__init__`, we should either accept a host/port or a `service`.

emontnemery: The wrapper is dumb, but without it pylint complains `initialize_connection()` was too long. OK to silence that?

Ok, will fix `__init__` as you suggest.

pychromecast/socket_client.py
emontnemery: Good point.

pychromecast/socket_client.py
balloob: yeah silence the pylint error

emontnemery: OK, fixed.

pychromecast/discovery.py
emontnemery: Fixed.

PR #268, opened 2019-02-15
Exponential backoff of mdns service resolution retries

emontnemery:
Exponential backoff of mdns service resolution retries since each resolution request may generate a large number of packets with detrimental side effects.

Also:
- Connect to IP, if known, instead of to host in case mdns name resolution is not enabled
- Improve debug prints to include friendly name if known

See home-assistant/home-assistant#21075

### DIFF
--- a/pychromecast/socket_client.py
+++ b/pychromecast/socket_client.py
@@ -174,6 +174,7 @@ class SocketClient(threading.Thread):
         self._force_recon = False
 
         self.cast_type = cast_type
+        self.fn = None
         self.tries = tries
         self.timeout = timeout or TIMEOUT_TIME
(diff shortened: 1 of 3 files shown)


pychromecast/socket_client.py
```diff
@@ -252,35 +263,40 @@ def initialize_connection(self):  # pylint:disable=too-many-statements
                     # Resolve the service name. If service is None, we're
                     # connecting directly to a host name or IP-address
                     if service:
+                        if now < retry['next_retry']:
```
balloob: Why check this here instead of before creating a new socket and outside of the try?

emontnemery: Hmm, you're right.. Will fix.

PR #269, opened 2019-02-16
Improve exponential backoff of mdns service resolution retries

emontnemery:
Improve exponential backoff if mdns service resolution introduced in #268 by also backing off if service resolution fails.
Also, small do not issue a `remove_service` callback if the service is unknown.


emontnemery:
@balloob This should be in HA 0.88

PR #271, opened 2019-03-01
Do not automatically start worker thread and connect in Chromecast constructor

emontnemery:
Do not automatically start worker thread or connect in `Chromecast` constructor.

This is a breaking change, and the user will now have to call either of:
- `Chromecast.start()`: Start the worker thread and connect to the chromecast device. Connection status will be reported through the listener registered in `Chromecast.register_connection_listener`.
- `Chromecast.wait()`: Wait for connection, this will also start the worker thread if it has not been started.
- `Chromecast.connect()`: Connect to the chromecast. This must only be called if the worker thread has not been started. Connection status will be reported through the listener registered in `Chromecast.register_connection_listener`.

Background:
The automatic connect in the constructor meant that the constructor would hang forever if the number of retries was unlimited and the chromecast could not be found.
It was also a bit unnatural to start the worker thread in the constructor.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -39,6 +39,8 @@ How to use
     ['Dev', 'Living Room', 'Den', 'Bedroom']
 
     >> cast = next(cc for cc in chromecasts if cc.device.friendly_name == "Living Room")
+    >> # Start worker thread
+    >> cast.start()
     >> # Wait for cast device to be ready
     >> cast.wait()
     >> print(cast.device)
--- a/README.rst
+++ b/README.rst
@@ -39,9 +39,7 @@ How to use
     ['Dev', 'Living Room', 'Den', 'Bedroom']
 
     >> cast = next(cc for cc in chromecasts if cc.device.friendly_name == "Living Room")
-    >> # Start worker thread
-    >> cast.start()
-    >> # Wait for cast device to be ready
+    >> # Start worker thread and wait for cast device to be ready
     >> cast.wait()
     >> print(cast.device)
     DeviceStatus(friendly_name='Living Room', model_name='Chromecast', manufacturer='Google Inc.', uuid=UUID('df6944da-f016-4cb8-97d0-3da2ccaa380b'), cast_type='cast')
--- a/examples/blocking.py
+++ b/examples/blocking.py
@@ -19,6 +19,7 @@ if len(casts) == 0:
     print("No Devices Found")
     exit()
 cast = casts[0]
+cast.start()
 
 print()
 print(cast.device)
--- a/examples/dashcast_blocking.py
+++ b/examples/dashcast_blocking.py
@@ -21,6 +21,7 @@ if len(casts) == 0:
     exit()
 
 cast = casts[0]
+cast.start()
 
 d = dashcast.DashCastController()
 cast.register_handler(d)
--- a/examples/non_blocking.py
+++ b/examples/non_blocking.py
@@ -20,6 +20,7 @@ def your_main_loop():
     t = 1
     cast = None
     def callback(chromecast):
+        chromecast.connect()
         nonlocal cast
         cast = chromecast
         stop_discovery()
--- a/examples/simple_listener_example.py
+++ b/examples/simple_listener_example.py
@@ -30,6 +30,7 @@ class StatusMediaListener:
 chromecasts = pychromecast.get_chromecasts()
 chromecast = next(cc for cc in chromecasts
                   if cc.device.friendly_name == "Living Room Speaker")
+chromecast.start()
 
 listenerCast = StatusListener(chromecast.name, chromecast)
 chromecast.register_status_listener(listenerCast)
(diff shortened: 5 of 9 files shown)


balloob:
When do we want to connect right away? Could we just do the connect call only inside the discover methods and not do any I/O in the constructor? (which shouldn't do I/O really)

emontnemery:
@balloob sure, but I guess this library is used by many people. Do you think it's ok to change the behaviour?

balloob:
Ah right, now I remember the usage. It is so that when you instantiate, you know if it works or not. I suggest we make this a breaking change and a major bump. Require people to manually call `initialize_connection`

emontnemery:
> Require people to manually call initialize_connection

`initialize_connection` is automatically called if we're disconnected from the cast, do you suggest maintaining the connection is now the user's responsibility? Or just the initial connect? If the latter, I think the call should rather be to start the worker thread.

balloob:
ah good one. Maybe we can do that?

emontnemery:
> Maybe we can do that?

I'm not sure what _that_ means in this case, do you just suggest ro remove this:
https://github.com/balloob/pychromecast/blob/ec0775a5aab56378f0761be01661fbb5fa74119e/pychromecast/__init__.py#L214-L215
And the user has to make that call instead? I'm not sure if it's very useful.

What may be useful though is to make the change in this PR, i.e. `defer_connect=True` the default and only supported mode?

balloob:
Yes, let's make it the only supported mode. We will do a major version bump. We should make sure we update our examples to show how existing code can be updated

README.rst
```diff
@@ -39,6 +39,8 @@ How to use
     ['Dev', 'Living Room', 'Den', 'Bedroom']
 
     >> cast = next(cc for cc in chromecasts if cc.device.friendly_name == "Living Room")
+    >> # Start worker thread
+    >> cast.start()
```
balloob: Thinking here, do you think that we can automatically call `start` inside `wait`, if not started yet ? That way it's a less breaking change.

balloob:
Could you write a message describing the breaking change for the release post?

emontnemery:
Sure, where do I write it?

balloob:
As a comment here :D

emontnemery:
OK, I updated the issue description to be clearer.

balloob:
Released as 3.0.0

strunker:
Struggling a bit getting 3.0 working again, for now going to revert to 2.4 or 2.5

Running through the simple example on the landing page I get the below:
`import pychromecast
chromecasts = pychromecast.get_chromecasts()
cast = next(cc for cc in chromecasts if cc.device.friendly_name == "Kitchen")
cast.connect()
mc = cast.media_controller
mc.play_media('http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4', 'video/mp4')`

Calling cast.wait() or cast.connect()
Results in the below error. 

  File "c:\python27\lib\threading.py", line 801, in __bootstrap_inner
    self.run()
  File "c:\python27\lib\site-packages\pychromecast\socket_client.py", line 423, in run
    self.initialize_connection()
  File "c:\python27\lib\site-packages\pychromecast\socket_client.py", line 252, in initialize_connection
    for service in self.services.copy():
AttributeError: 'list' object has no attribute 'copy'

Calling cast.start() results in the below error:

  File "C:\Python27\Echo\test.py", line 466, in <module>
    mc.play_media('http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4', 'video/mp4')
  File "c:\python27\lib\site-packages\pychromecast\controllers\media.py", line 480, in play_media
    callback_function=app_launched_callback)
  File "c:\python27\lib\site-packages\pychromecast\socket_client.py", line 925, in launch_app
    self.update_status(lambda response:
  File "c:\python27\lib\site-packages\pychromecast\socket_client.py", line 916, in update_status
    callback_function=callback_function_param)
  File "c:\python27\lib\site-packages\pychromecast\controllers\__init__.py", line 84, in send_message
    self.namespace, data, inc_session_id, callback_function)
  File "c:\python27\lib\site-packages\pychromecast\socket_client.py", line 721, in send_platform_message
    inc_session_id, callback_function_param)
  File "c:\python27\lib\site-packages\pychromecast\socket_client.py", line 668, in send_message
    self._ensure_channel_connected(destination_id)
  File "c:\python27\lib\site-packages\pychromecast\socket_client.py", line 758, in _ensure_channel_connected
    no_add_request_id=True)
Exception in thread Thread-1:
  File "c:\python27\lib\threading.py", line 801, in __bootstrap_inner
    self.run()
  File "c:\python27\lib\site-packages\pychromecast\socket_client.py", line 423, in run
    self.initialize_connection()
  File "c:\python27\lib\site-packages\pychromecast\socket_client.py", line 252, in initialize_connection
    for service in self.services.copy():
AttributeError: 'list' object has no attribute 'copy'
  File "c:\python27\lib\site-packages\pychromecast\socket_client.py", line 715, in send_message

    str(self.port) + " is connecting...")
pychromecast.error.NotConnected: Chromecast 10.1.1.12:8009 is connecting...

PR #275, opened 2019-03-14
Fix Spotify playback

fondberg:
Spotify now requires the expiresIn token in the message.
Tested to work with Harman Kardon Google Home smart speaker and Chromecast stick v1.

### DIFF
--- a/pychromecast/controllers/spotify.py
+++ b/pychromecast/controllers/spotify.py
@@ -41,7 +41,8 @@ class SpotifyController(BaseController):
         def callback():
             """Callback function"""
             self.send_message({"type": TYPE_STATUS,
-                               "credentials": self.access_token})
+                               "credentials": self.access_token,
+                               "expiresIn": 3600})
 
         self.launch(callback_function=callback)
 


emontnemery:
@fondberg How do you test? I did some testing with `spotify_example.py` and it doesn't really seem to work, with or without your change.

fondberg:
@emontnemery I tested with the spotify_example.py as well.

fondberg:
Just tested on other devices and it works. Are you sure your test passed this:
```
if cast.name == CAST_NAME:
```

emontnemery:
There was a problem with pyspotify also, the version available from pip is outdated, I had to install directly from github instead.

What is the recommended way of installing?

emontnemery:
@fondberg Thanks, it is indeed working!

pychromecast/controllers/spotify.py
```diff
@@ -41,7 +41,8 @@ def launch_app(self):
         def callback():
             """Callback function"""
             self.send_message({"type": TYPE_STATUS,
-                               "credentials": self.access_token})
+                               "credentials": self.access_token,
+                               "expiresIn": 3600})
```
balloob: So `expiresIn` should indicate when the token expires. This value is in seconds. It is part of the access token ad refresh token responses as per the OAuth2 spec.

```python
# Refresh token response
{
    "access_token": "ABCDEFGH",
    "expires_in": 1800,
    "refresh_token": "IJKLMNOPQRST",
    "token_type": "Bearer"
}

# Access token response
{
    "access_token": "ABCDEFGH",
    "expires_in": 1800,
    "token_type": "Bearer"
}
```

Although this PR will allow the frontend to work, it will not function correctly, and it will drop the token in 1 hour?

balloob: I think the correct solution would be to:
 - allow expires_in to be passed in. Default expires_in to the default expires_in that Spotify returns (although this might cause weird issues in the future).
 - automatically adjust the value if we send this message later than it was passed in
 - throw an error if the token has expires

fondberg:
Agree that oj would be better. I'll try to create a PR with using the token expiration from the response when I get some time

PR #278, opened 2019-03-17
WIP - Add multizone controller

### LINKED ISSUE #111: Chromecast Audio status when playing as member of a group
When a Chromecast Audio is playing media as a member of a group, the returned MEDIA_STATUS is empty:
`{u'status': [], u'type': u'MEDIA_STATUS', u'requestId': 3}`

What would be the correct way reliably know if a particular device is playing, including when it's acting as a member of a group?
It would be possible to loop over all groups of which the device is a member, but:
- How to get a list of groups of which the device is a member?
- What if several groups, of which a particular device is a member, are playing at the same time?

Edit:
Partial solution:
In each group, one Chromecast will be elected 'leader', while the remaining players in the same group will be 'followers'.

When the **leader** is playing or paused, it will report an app_id of 'MultizoneLeader' and a 'session_id' which is same as the 'session_id' of the group it is leading. Hence, state of the leader can be deduced from the `media_controller.status.player_state` state of the matching group.

The **followers** will report an app_id of 'MultizoneFollower'.
Doing:
`requests.get('http://' + follower_ip + ':8008/setup/eureka_info?params=multizone')`
will return, among other group related settings, a list of groups in which the follower is a member.
The state of the follower can be guessed by comparing status.status_text and status.display_name between the follower and groups in which it is a member.

There is probably a more straight forward way to do this.

emontnemery:
Fixes #111

### DIFF
--- a/pychromecast/controllers/multizone.py
+++ b/pychromecast/controllers/multizone.py
@@ -1,0 +1,85 @@
+"""
+Controller to monitor audio group members.
+"""
+import json
+import logging
+import requests
+
+from . import BaseController
+from .. import discovery
+
+_LOGGER = logging.getLogger(__name__)
+
+MESSAGE_TYPE = "type"
+MULTIZONE_NAMESPACE = "urn:x-cast:com.google.cast.multizone"
+TYPE_DEVICE_ADDED = "DEVICE_ADDED"
+TYPE_DEVICE_UPDATED = "DEVICE_UPDATED"
+TYPE_DEVICE_REMOVED = "DEVICE_REMOVED"
+TYPE_GET_STATUS = "GET_STATUS"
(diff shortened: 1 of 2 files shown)


pychromecast/controllers/multizone.py
```diff
@@ -0,0 +1,84 @@
...
+                self._callback.multizone_member_added(uuid)
+
+    def _remove_member(self, uuid):
+        name = self._members.pop(uuid, '<Unknown>')
+        _LOGGER.debug("(%s) Removed member %s(%s), members: %s",
+                      self._uuid, uuid, name, self._members)
+        if self._callback:
+            self._callback.multizone_member_removed(uuid)
+
+    def get_members(self):
```
emontnemery: Could we have this called automatically when the socket_client connects?

PR #283, opened 2019-03-25
Add MultizoneManager

emontnemery:
Refactor MultizoneController
Add MultizoneManager which helps keep track of which casts are members of which groups
Add new Dial method go get Multizone information

### DIFF
--- a/pychromecast/controllers/multizone.py
+++ b/pychromecast/controllers/multizone.py
@@ -4,7 +4,7 @@ Controller to monitor audio group members.
 import logging
 
 from . import BaseController
-from .media import MediaController, MediaStatus
+from .media import MediaController
 from ..socket_client import (
     CONNECTION_STATUS_CONNECTED, CONNECTION_STATUS_DISCONNECTED,
     CONNECTION_STATUS_LOST)
--- a/pychromecast/controllers/multizone.py
+++ b/pychromecast/controllers/multizone.py
@@ -2,11 +2,11 @@
 Controller to monitor audio group members.
 """
 import logging
-from uuid import UUID  # pylint: disable=unused-import
+from uuid import UUID  # noqa: F401,E501 pylint: disable=unused-import
 import threading
 
 from . import BaseController
-from .media import MediaController, MediaStatus  # pylint: disable=unused-import
+from .media import MediaController, MediaStatus  # noqa: F401,E501 pylint: disable=unused-import
 from ..socket_client import (
     CONNECTION_STATUS_CONNECTED, CONNECTION_STATUS_DISCONNECTED,
     CONNECTION_STATUS_LOST)
@@ -118,7 +118,7 @@ class MultizoneManager:
         group = self._groups.pop(group_uuid, None)
         # Inform all group members that they are no longer members
         if group is not None:
-            group['listener']._mz.reset_members()  # pylint: disable=protected-access
+            group['listener']._mz.reset_members()  # noqa: E501 pylint: disable=protected-access
         with self._lock:
             for member in self._casts.values():
                 member['groups'].discard(group_uuid)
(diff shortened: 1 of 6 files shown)


pychromecast/controllers/multizone.py
```diff
@@ -2,27 +2,173 @@
 Controller to monitor audio group members.
 """
 import logging
+from uuid import UUID  # pylint: disable=unused-import
```
emontnemery: Why is flake and pylint complaining about this import?

pychromecast/controllers/multizone.py
```diff
@@ -2,27 +2,173 @@
 Controller to monitor audio group members.
 """
 import logging
+from uuid import UUID  # noqa: F401,E501 pylint: disable=unused-import
```
emontnemery: Why is flake and pylint complaining about this import?

balloob: Because you're not using it ?

pychromecast/controllers/multizone.py
```diff
@@ -2,27 +2,173 @@
...
+        # the casts' socket_client thread
+        self._casts = {}
+        self._groups = {}
+        self._lock = threading.Lock()
+
+    def add_multizone(self, group_cast):
+        """ Start managing a group """
+        lock = self._lock
+
+        class Listener:
```
balloob: Pull class definition out of method.

pychromecast/controllers/multizone.py
```diff
@@ -2,27 +2,173 @@
...
+    def __init__(self):
+        # Protect self._casts because it will be accessed from callbacks from
+        # the casts' socket_client thread
+        self._casts = {}
+        self._groups = {}
+        self._lock = threading.Lock()
+
+    def add_multizone(self, group_cast):
+        """ Start managing a group """
+        lock = self._lock
```
balloob: oh no, locks ! Can we do without?

pychromecast/controllers/multizone.py
```diff
@@ -2,27 +2,173 @@
...
+                                listener.multizone_new_cast_status(
+                                    self._group_uuid, cast_status)
+
+            def new_media_status(self, media_status):
+                """Handle reception of a new MediaStatus."""
+                with lock:
+                    casts = self._casts
+                    group_members = self._mz.members
+                    for member_uuid in group_members:
+                        if member_uuid in casts:
```
balloob: May I suggest some guard clauses:

```python
if member_uuid not in casts:
    continue

for …
```

pychromecast/controllers/multizone.py
```diff
@@ -2,27 +2,173 @@
...
+                    if member_uuid not in casts:
+                        casts[member_uuid] = {'listeners': [],
+                                              'groups': set()}
+                    casts[member_uuid]['groups'].add(self._group_uuid)
+                    for listener in casts[member_uuid]['listeners']:
+                        listener.added_to_multizone(self._group_uuid)
+
+            def multizone_member_removed(self, member_uuid):
+                """Handle removed audio group member."""
+                with lock:
```
balloob: Can we do it without locks?

pychromecast/controllers/multizone.py
emontnemery: We need to protect self._casts because it will be accessed from callbacks from the casts' socket_client threads.
It could be modified such that the lock is only held when accessing the dict though, or remove the lock and rely on the GIL:
> This simplifies the CPython implementation by making the object model (including critical built-in types such as dict) implicitly safe against concurrent access.

pychromecast/controllers/multizone.py
```diff
@@ -31,22 +161,44 @@ def _add_member(self, uuid, name):
...
+            for listener in self._status_listeners:
+                listener.multizone_member_added(uuid)
 
     def _remove_member(self, uuid):
         name = self._members.pop(uuid, '<Unknown>')
         _LOGGER.debug("(%s) Removed member %s(%s), members: %s",
                       self._uuid, uuid, name, self._members)
-        if self._callback:
-            self._callback.multizone_member_removed(uuid)
+        for listener in self._status_listeners:
```
balloob: let's make a copy to prevent mutated while iteration:

```suggestion
        for listener in list(self._status_listeners):
```

emontnemery: Fixed, along with a couple more same.

czesiu:
@emontnemery do you know how to create dynamic_group? Is there a option in google home app? I see in the code there is a field called 'dynamic_groups' in multizone status and that's why I'm asking.

PR #284, opened 2019-03-29
Add google home mini as audio device

Tsjippy:


Tsjippy:
A home mini is now show as a cast device, but it is only an audio device. Please merge this.

balloob:
> Please merge this

You didn't add a description, so I did not consider it worth my time. Don't just ask people to merge things. Code needs to be reviewed and it needs to be tested that it works.

Tsjippy:
Ok thanks! Sorry that I forgot about that, next time I will do.

PR #293, opened 2019-05-08
Example on how to cast to group

phr0z:


marcelveldt:
There is cleaner way to do this:
```

class SpController(SpotifyController):
    """ Controller to interact with Spotify namespace. """
    def receive_message(self, message, data):
        """ handle the auth flow and active player selection """
        if data['type'] == 'setCredentialsResponse':
            self.send_message({'type': 'getInfo', 'payload': {}})
        if data['type'] == 'setCredentialsError':
            self.device = None
        if data['type'] == 'getInfoResponse':
            self.device = data['payload']['deviceID']
            self.is_launched = True
        return True
```

controller.device will contain the correct spotify device id after launch, works with all players, also grouped ones.

emontnemery:
@marcelveldt interesting!

PR #294, opened 2019-05-11
Update spotify controller and spotify example to handle casting to audio groups

emontnemery:


balloob:
Did you see  #293 ? :)

fondberg:
Looks interesting. I will test this as soon as I have a chance

examples/spotify_example.py
```diff
@@ -6,46 +6,147 @@
...
+                break
+
+    if not device_id:
+      print('Could not find matching member of audio group')
+      print('Known devices: {}'.format(devices_available['devices']))
+      print('Members of group "{}": {}'.format(args.cast, mz.members))
+      sys.exit(1)
+
+# Start playback
+client.start_playback(device_id=device_id, uris=[args.track])
```
fondberg: It would be nice to support playlists as well like
```
if args.uri.find('track') > 0:
    client.start_playback(device_id=spotify_device_id, uris=[args.uri])
else:
    client.start_playback(device_id=spotify_device_id, context_uri=args.uri)
```
And change args accordingly:
```
parser.add_argument('--uri', help='Spotify uri (default: "%(default)s")',
                    default="spotify:track:3Zwu2K0Qa5sT6teCCHPShP")
```

examples/spotify_example.py
```diff
@@ -6,46 +6,147 @@
...
+    mz_listener = MzListener()
+    mz.register_listener(mz_listener)
+    cast.register_handler(mz)
 
-    client = spotipy.Spotify(auth=access_token)
+    # Register a connection listener, when connected, we can poll group members
+    cast.register_connection_listener(ConnListener(mz))
+    
+# Wait for connection to the chromecast
+cast.wait()
```
fondberg: Is this the same for a group?

emontnemery:
@fondberg this https://github.com/balloob/pychromecast/pull/293#issuecomment-491532612 might be the way to do it, if it works there's no need to go through the multizone hoops

fondberg:
@emontnemery it works great. My small suggestion can be ignored if you want and you can merge this. Can you make a release when this is in?

emontnemery:
@fondberg I added your small suggestion, and also tried to improve the error handling to fail if there is a credential error.
Maybe this should be replaced with a wait on a future instead of looping on a boolean though:
https://github.com/balloob/pychromecast/blob/68012f0a7cf3ff874d37932137b25e3afe818d8d/pychromecast/controllers/spotify.py#L71-L74

fondberg:
LGTM. Which thread would send the event (not a master on pychromecast... yet...)?

emontnemery:
It will be sent by the SocketClient worker thread which is handling communication with the cast.

PR #295, opened 2019-05-31
Seek to a specified point in time via the Plex controller

tking2:
Adds the ability to seek to a specified time while playing media through Plex.


balloob:
Thanks!

PR #299, opened 2019-06-26
Use the callback if its provided.

Hellowlol:


pychromecast/controllers/__init__.py
```diff
@@ -73,7 +73,7 @@ def send_message(self, data, inc_session_id=False,
         if not self.target_platform and \
            self.namespace not in self._socket_client.app_namespaces:
             if self.supporting_app_id is not None:
-                self.launch()
+                self.launch(callback_function=callback_function)
```
balloob: This means the callback will be called twice, which doesn't seem correct.

Hellowlol: Where else is this executed?  I have been writing a controller new controller to plex so we can start Playback but it seems to fail sometime if the app isn’t loaded from before. I assumed this was because the callback wasn’t used, as the app opens but nothing else happens.

balloob: Line 84.

Hellowlol:
Closing this, as i was mistaken. The callback is executed but we only block until the app is launched, but until the callback is done. Thanks!

PR #300, opened 2019-07-01
Better controller for plex.

Hellowlol:
Add a enchanted Plex controller with additional methods.

There is also a plexapi controller that have some additional features, like change subtitle, audio etc. It isnt the cleanest method to do this but i didnt see any other way to add support for this.

Note:
This pr will not require that plexapi is a requirement pychromecast. All methods in PlexController can be used without plexapi, but using it make the usage alot easier and there is less stuff to set manually.

### DIFF
--- a/pychromecast/controllers/plex.py
+++ b/pychromecast/controllers/plex.py
@@ -9,33 +9,51 @@ from urllib.parse import urlparse
 
 from . import BaseController
 
-MESSAGE_TYPE = 'type'
-STREAM_TYPE_UNKNOWN = 'UNKNOWN'
-STREAM_TYPE_BUFFERED = 'BUFFERED'
-STREAM_TYPE_LIVE = 'LIVE'
+MESSAGE_TYPE = "type"
+STREAM_TYPE_UNKNOWN = "UNKNOWN"
+STREAM_TYPE_BUFFERED = "BUFFERED"
+STREAM_TYPE_LIVE = "LIVE"
 SEEK_KEY = "currentTime"
-TYPE_PLAY = 'PLAY'
-TYPE_PAUSE = 'PAUSE'
-TYPE_STOP = 'STOP'
-TYPE_STEPFORWARD = 'STEPFORWARD'
-TYPE_STEPBACKWARD = 'STEPBACK'
-TYPE_PREVIOUS = 'PREVIOUS'
-TYPE_NEXT = 'NEXT'
-TYPE_LOAD = 'LOAD'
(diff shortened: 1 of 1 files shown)


Hellowlol:
If someone please would tell me how i can make travis happy? please let me know so i can rebase this thing.

pychromecast/controllers/plex.py
```diff
@@ -1,40 +1,538 @@
...
+    @property
+    def status(self):
+        """Get the chromecast playing status.
+
+        Returns:
+            pychromecast.controllers.media.MediaStatus: Slightly modified status with patched
+                                                        method for episode_title.
+        """  # noqa
+
+        @property
```
balloob: I don't think that this should be defined inside `status` ?

pychromecast/controllers/plex.py
```diff
@@ -1,40 +1,538 @@
...
+
+            self._send_cmd(
+                msg,
+                namespace="urn:x-cast:com.google.cast.media",
+                inc_session_id=True,
+                inc=False,
+            )
+        else:
+            self.logger.debug(
+                "Cant reset teh stream as _last_play_msg "
```
balloob:
```suggestion
                "Cant reset the stream as _last_play_msg "
```

pychromecast/controllers/plex.py
```diff
@@ -1,40 +1,538 @@
...
+
+    def disconnect(self, timeout=None, blocking=True):
+        """Disconnect the controller."""
+        self._socket_client.disconnect()
+        if blocking:
+            self.join(timeout=timeout)
+
+
+# pylint: disable=too-many-public-methods
+class PlexApiController(PlexController):
```
balloob: Why does this class exist? `self.pms` is not being used.

Hellowlol: Its being used inside _get_current_media, I have separated this a in its own class as I haven't found another way to to do the public methods in this class without plexapi.

pychromecast/controllers/plex.py
Hellowlol: Do you want me to add this to MediaStatus or just define in outside the property?

Hellowlol:
Is there anything else i need to do to get this merged?

balloob:
Looks good! Thanks

Nojahhh:
Thanks for a great update on the PlexController @Hellowlol.

Tried this out and its working great with plexapi.

Hellowlol:
@Nojahhh Thanks! It's nice to know others could benefit. I needed chromecast support for my own project :)

PR #305, opened 2019-08-11
Fix failure to reconnect after device reboot

soreau:
After chromecast reboot, the device cannot be reconnected in
response to CONNECTED status message with restarting the
application. This reinitializes the session to facilitate this purpose.

### DIFF
--- a/pychromecast/dial.py
+++ b/pychromecast/dial.py
@@ -13,9 +13,6 @@ XML_NS_UPNP_DEVICE = "{urn:schemas-upnp-org:device-1-0}"
 
 FORMAT_BASE_URL = "http://{}:8008"
 
-CC_SESSION = requests.Session()
-CC_SESSION.headers['content-type'] = 'application/json'
-
 # Regular chromecast, supports video/audio
 CAST_TYPE_CHROMECAST = 'cast'
 # Cast Audio device, supports only audio
@@ -37,8 +34,10 @@ _LOGGER = logging.getLogger(__name__)
 
 def reboot(host):
     """ Reboots the chromecast. """
-    CC_SESSION.post(FORMAT_BASE_URL.format(host) + "/setup/reboot",
-                    data='{"params":"now"}', timeout=10)
+    headers = {'content-type': 'application/json'}
+
+    requests.post(FORMAT_BASE_URL.format(host) + "/setup/reboot",
+                  data='{"params":"now"}', headers=headers, timeout=10)
 
 
 def _get_status(host, services, zconf, path):
@@ -57,7 +56,9 @@ def _get_status(host, services, zconf, path):
                 _LOGGER.debug("Resolved service %s to %s", service, host)
                 break
 
-    req = CC_SESSION.get(FORMAT_BASE_URL.format(host) + path, timeout=10)
+    headers = {'content-type': 'application/json'}
+
+    req = requests.get(FORMAT_BASE_URL.format(host) + path, headers=headers, timeout=10)
 
     req.raise_for_status()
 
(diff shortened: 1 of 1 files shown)


pychromecast/dial.py
```diff
@@ -57,6 +57,10 @@ def _get_status(host, services, zconf, path):
                 _LOGGER.debug("Resolved service %s to %s", service, host)
                 break
 
+    # Reinitialize the session in case the device was rebooted
+    CC_SESSION = requests.Session()
```
skorokithakis: All this does is create a local variable (hence the CI errors, I believe). You need `global CC_SESSION` above this line to reassign the global.

soreau: Thanks @skorokithakis, I updated the patch

skorokithakis: That looks much better, the CI failure now looks like a linter problem, to me. At least, I don't see how this can be done without using the `global` statement.

balloob: I think that we can get rid of the CC_SESSION completely. There is no reason to keep one around ?

soreau: @balloob Hi, CC_SESSION is used for .get() in _get_status() and .post() in reboot(). What would be the alternative if there is no CC_SESSION for .get() and.post()?

soreau: @balloob [This also works](http://ix.io/1RBj) which changes CC_SESSION to a local variable that is initialized with the session each time it is used.

balloob: Well I was thinking why wouldn't we just use `requests.get` directly ? What is in the new session that fixes it?

soreau: If I try to print CC_CESSION before and after it's set to `requests.Session()` and it says `UnboundLocalError: local variable 'CC_SESSION' referenced before assignment` for the before statement, in _get_status. If I try `requests.get` it silently fails. .get() is a method on `request.Session()`, not `request`. `request.Session().get` also fails. It needs a variable to assign `request.Session()`, then needs `.headers['content-type'] = 'application/json'` called on it, then it works. This requires a variable unless you blindly call `request.Session()` twice, once to set json and once for `get/post`, but this also fails anyway. As far as I can tell, a variable is needed regardless.

balloob: requests.get can have a headers dictionary be passed in. https://2.python-requests.org/en/master/user/quickstart/#custom-headers

soreau: @balloob Yes that does work, thanks. I have updated the patch.

balloob:
Ok to merge after you fix the formatting so that Travis passes.

soreau:
> Ok to merge after you fix the formatting so that Travis passes.

I fixed the formatting but now there's another error I don't understand.

balloob:
```
$ flake8 --exclude cast_channel_pb2.py,authority_keys_pb2.py,logging_pb2.py pychromecast

pychromecast/dial.py:61:80: E501 line too long (88 > 79 characters)
```

soreau:
Thanks, I would have used black but it would reformat the entire file..

balloob:
I am open for a PR switching this repo to black and enforced by CI 👍

soreau:
Thanks for the help with this. How can I find out in what earliest release it might appear?

soreau:
@balloob For reference, and as a shameless self promotion, I present to you, [my control gui](https://github.com/soreau/catt-qt).

balloob:
I'll do a release tomorrow. 

Nice work on the GUI!

soreau:
@balloob Awesome, you rock. Thanks for pychromecast!

balloob:
3.2.3 released.

PR #312, opened 2019-09-10
Drop support for Py34, Py35

balloob:
Drop support for Python 3.4 and Python 3.5. They are old, even Debian Stable is running Py 3.7.


p1r473:
My TinkerBoard only supports Debian Stretch which is running Python 3.5

emontnemery:
@p1r473 please note this PR was merged September 2019.

PR #325, opened 2019-11-10
Add support for casting with Yle Areena app.

Eerovil:
https://areena.yle.fi/

This is Finlands government-run streaming service. The service uses the Kaltura player for video streaming, and thus fetching media ID's is a bit difficult. I've worked around this by just opening a page and running the javascript, to get an "Entry ID".


pychromecast/controllers/yleareena.py
```diff
@@ -0,0 +1,82 @@
...
+    # pylint: disable=useless-super-delegation
+    def __init__(self):
+        super(YleAreenaController, self).__init__()
+
+        self.app_id = APP_YLEAREENA
+
+    # pylint: disable=too-many-arguments
+    def play_areena_media(
+        self,
+        entry_id="",
```
balloob: This is required, don't give it a default value.

emontnemery:
@Eerovil This has been sitting for some time, do you consider it ready for merge?

Eerovil:
I'm not sure, I'd like some input from a tester perhaps. But then again, merging won't hurt anybody. 

Personally I'm using direct media urls to cast Areena, so this feature isn't getting much use from me.

Eerovil:
I've found a better way for fetching id's. Now I'm happy with the pull request!

emontnemery:
Implemented in #359

PR #334, opened 2020-01-17
Correct `METADATA_TYPE` in media.py

jparthum:
Per the [Google Developer Docs](https://developers.google.com/cast/docs/reference/messages#MediaInformation), `MovieMediaMetadata` should be `metadata: 1` and `TvShowMediaMetadata` should be `metadata: 2`.


balloob:
Thanks!

Issue #332, opened 2020-01-02
I get to cast.wait() and then get Failed to connect


stefan52a:
```
>> import time
>> import pychromecast

>> chromecasts = pychromecast.get_chromecasts()
>> [cc.device.friendly_name for cc in chromecasts]
['Kantoor']

>> cast = next(cc for cc in chromecasts if cc.device.friendly_name == "Kantoor")
>> # Start worker thread and wait for cast device to be ready
>> cast.wait()
```

then get

```
Casting to: Kantoor
 
[record[a,in-unique,7c5d0363-398a-f3e9-f9be-3c5c198063fa.local.]=120/119,10.57.4.8:8009] Failed to connect, retrying in 5.0s

```

What is wrong? Tried same pc and chromecast (which is a google nest hub BTW) in the same WiFi.
Note: I use LAN range 10.57.4.x

kcowgill:
Same; trying to connect to a Google Home on the same subnet.


`
$ python
Python 3.6.5 |Anaconda, Inc.| (default, Apr 26 2018, 08:42:37)
[GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import time
>>> import pychromecast
>>> chromecasts = pychromecast.get_chromecasts()
>>> [cc.device.friendly_name for cc in chromecasts]
['Upstairs Speaker', 'Basement Speaker', 'Living Room Home']
>>> cast = next(cc for cc in chromecasts if cc.device.friendly_name == "Living Room Home")
>>> cast.wait()
[record[a,in-unique,b6c5b718-6b2d-8767-c8dc-a7263f0ce36f.local.]=120/119,192.168.1.79:8009] Failed to connect, retrying in 5.0s
`

kcowgill:
After some poking I'm thinking there's something up with discovery/`get_chromecasts()`...

```
$ python
Python 3.6.5 |Anaconda, Inc.| (default, Apr 26 2018, 08:42:37)
[GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import pychromecast
>>> cast = pychromecast.Chromecast('192.168.1.79')
>>> cast.wait()
>>> cast.status
[...]

PR #337, opened 2020-01-20
Get ipaddress the same way as code examples in zeroconf

fondberg:
zero-conf 0.24.4 changes the way the information from `zconf.cache.entries_with_name` is returned. According to @jstasiak  the current way is [unsupported](https://github.com/balloob/pychromecast/issues/332#issuecomment-576258110). 
This PR gets the ipaddress the official way (https://github.com/jstasiak/python-zeroconf/blob/master/examples/browser.py#L22).

### DIFF
--- a/pychromecast/discovery.py
+++ b/pychromecast/discovery.py
@@ -67,8 +67,8 @@ class CastListener:
                 return value
             return value.decode("utf-8")
 
-        addresses = ["%s" % socket.inet_ntoa(addr) for addr in service.addresses]
-        host = addresses[0] if len(addresses) > 0 else service.server
+        addresses = service.parsed_addresses()
+        host = addresses[0] if addresses else service.server
 
         model_name = get_value("md")
         uuid = get_value("id")
--- a/pychromecast/discovery.py
+++ b/pychromecast/discovery.py
@@ -43,6 +43,7 @@ class CastListener:
 
     def add_service(self, zconf, typ, name):
         """ Add a service to the collection. """
+        import socket
         service = None
         tries = 0
         _LOGGER.debug("add_service %s, %s", typ, name)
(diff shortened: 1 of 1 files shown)


jstasiak:
Oh, forgot to mention one thing - if you want to convert `ServiceInfo.addresses` to strings you may want to use `ServiceInfo.parsed_addresses()` method, it'll do the conversion for you (https://github.com/jstasiak/python-zeroconf/blob/29432bfffd057cf4da7636ba0c28c9d8a7ad4357/zeroconf/__init__.py#L1611).

fondberg:
Thanks, I'll try that tomorrow!

fondberg:
I looked into it a bit and the question and to not introduce new functionality for pychromecast it needs to restrict to IpV4 only. I leave this up to the review of the PR to decide:
```
addresses = service.parsed_addresses(zeroconf.IPVersion.V4On)
```

balloob:
I'm fine with adding ipv6 support. 👍

pychromecast/discovery.py
```diff
@@ -67,8 +67,8 @@ def get_value(key):
                 return value
             return value.decode("utf-8")
 
-        ips = zconf.cache.entries_with_name(service.server.lower())
-        host = repr(ips[0]) if ips else service.server
+        addresses = ["%s" % socket.inet_ntoa(addr) for addr in service.addresses]
+        host = addresses[0] if len(addresses) > 0 else service.server
```
balloob: To satisfy pylint:

```suggestion
        host = addresses[0] if addresses else service.server
```

fondberg: Will fix

fondberg:
@balloob I changed to use the helper method instead. I'm still using the first address in the returned list but an enhancement in the future would probably be to store the list of addresses in the `self.services[name]` dict.

fondberg:
We need to upload new release to pypi and bump version in home-assistant. Can you do a realiease and I can bump the version and in HA and do some tests on it?

balloob:
Already done both ;-)

fondberg:
You are on fire man 🚀🚀🚀

georgedpeterson:
Hello. I have this issue now with the most recent version. Any ideas? Thanks!

PR #348, opened 2020-04-14
Add helper function get_listed_chromecasts

emontnemery:
Add helper function `get_listed_chromecasts` and use it in examples.

This differs from `get_chromecasts` in that it attempts to discover listed chromecasts and then immediately returns.

### DIFF
--- a/examples/list_chromecasts.py
+++ b/examples/list_chromecasts.py
@@ -9,7 +9,7 @@ import time
 import pychromecast
 
 parser = argparse.ArgumentParser(
-    description="Example on how to use the Spotify Controller.")
+    description="Example on how to list chromecasts.")
 parser.add_argument('--show-debug', help='Enable debug log',
                     action='store_true')
 args = parser.parse_args()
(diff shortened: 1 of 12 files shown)


examples/list_chromecasts.py
```diff
@@ -0,0 +1,28 @@
...
+"""
+import argparse
+import logging
+import sys
+import time
+
+import pychromecast
+
+parser = argparse.ArgumentParser(
+    description="Example on how to use the Spotify Controller.")
```
balloob, line 12: description is old.

pychromecast/__init__.py
```diff
@@ -57,7 +59,10 @@ def _get_chromecast_from_host(
     )
 
 
-def _get_chromecast_from_service(
+_get_chromecast_from_host = get_chromecast_from_host  # pylint: disable=invalid-name
```
balloob, line 62: Why? And can you leave that as a comment.

PR #349, opened 2020-04-15
Remove blocking option from Chromecast

emontnemery:
Remove blocking option from Chromecast objects

Most operations were not blocking anyway and ReceiverController._block_till_launched would deadlock if the app did not launch.

This is a breaking change because `blocking` parameter is no longer accepted for several functions.


pychromecast/__init__.py
```diff
@@ -184,13 +169,7 @@ def get_chromecasts(
...
     :param callback: Callback which is triggerd for each discovered chromecast when
                      blocking = False.
-    :param blocking_app_launch: If set to true, start_app() will wait for the app
-                                to launch before returning. The default is to set
-                                blocking_app_launch = blocking.
     """
-    if blocking_app_launch is None:
-        blocking_app_launch = blocking
-
     if blocking:
```
balloob, line 173: Still a blocking here?

emontnemery: Right, this is a different `blocking`, I think it makes sense to keep it:
```
:param blocking: If True, returns a list of discovered chromecast devices.
                 If False, triggers a callback for each discovered chromecast,
                 and returns a function which can be executed to stop discovery.
```

PR #352, opened 2020-04-16
Speed up SocketClient shutdown

emontnemery:


pychromecast/socket_client.py
```diff
@@ -434,6 +435,11 @@ def connect(self):
     def disconnect(self):
         """ Disconnect socket connection to Chromecast device """
         self.stop.set()
+        try:
+            self.socketpair[1].send(b"x")
```
balloob, line 439: Can you add a comment on what is going on here ?

emontnemery: Sure, fixed

PR #359, opened 2020-05-03
Quick play & Support for Yle Areena, Supla

Eerovil:
This Pull request implements support for the home assistant service `cast_app`

https://github.com/home-assistant/core/pull/35114/

Note: this pull request also contains changes from PR #325 and additional Supla support


emontnemery:
> this pull request also contains changes from PR #325 and additional Supla support

Does it mean #325 is no longer needed?

Eerovil:
Yes, if this will be merged.

emontnemery:
Can you explain a bit the idea of the new `quick_play` function?

pychromecast/quick_play.py
```diff
@@ -0,0 +1,43 @@
...
+    elif app_name == "supla":
+        controller = SuplaController()
+        kwargs = {
+            "media_id": data.pop("media_id"),
+            "is_live": data.pop("extra2", None),
+        }
+    else:
+        raise NotImplementedError()
+
+    cast.wait()
```
emontnemery, line 38: We should not do a `cast.wait()` here:
- It's not suitable for Home Assistant because it blocks
- It's not compatible with running the SocketClient in a custom event loop instead of with its own thread.

Instead, the caller of `quick_play` should ensure the Chromecast object is in the correct state.

pychromecast/quick_play.py
```diff
@@ -0,0 +1,43 @@
...
+        vol.Required('app_name', default=""): cv.string,
+        vol.Required('data'): vol.Schema({
+            vol.Required("media_id"): cv.string,
+            vol.Optional("media_type"): cv.string,
+            vol.Optional("enqueue", default=False): cv.boolean,
+            vol.Optional("index"): cv.string,
+            vol.Optional("extra1"): cv.string,
+            vol.Optional("extra2"): cv.string,
+        }),
+    }
```
emontnemery, line 19: Do a proper docstring instead.

Eerovil:
So, the quick_play.quick_play command is supposed to be a direct listener to the HomeAssistant service cast_app.

Then, each Controller can also implement their own quick_play method, which is should contain everything needed to get the controller up and running a piece of media. (Sort of replacing the `examples/` directory).

Though, with most controllers, quick play is not possible to include in pychromecast since fetching chromecast-usable data requires app-specific dependencies. These can be implemented in the Cast Extensions repo.

pychromecast/quick_play.py
Eerovil: Could I use an app launch callback here instead?

pychromecast/quick_play.py
Eerovil: Yup, I'll document it.

I wish I could make the schema even better. The extra1 and extra2 keys are hard to keep track of, but I don't see another way of doing this.

pychromecast/quick_play.py
emontnemery: That should not be needed, `BaseController.send_message()` will automatically launch the necessary app and wait for it to be launched before sending the message:
https://github.com/home-assistant-libs/pychromecast/blob/7372f8d24209aad9b5f7f61f5988f28994a7ec60/pychromecast/controllers/__init__.py#L76-L81

pychromecast/controllers/yleareena.py
```diff
@@ -0,0 +1,88 @@
...
+    ):
+        """
+        Play media with the entry id "entry_id".
+        This value can be found by loading a page on Areena, e.g. https://areena.yle.fi/1-50097921
+        And finding the kaltura player which has an id of yle-kaltura-player3430579305188-29-0_whwjqpry
+        In this case the entry id is 0_whwjqpry
+        """
+
+        # pylint: disable=too-many-locals
+        def app_launched_callback():
```
emontnemery, line 36: This is not necessary, `BaseController.send_message()` will automatically launch the necessary app and wait for it to be launched before sending the message:
https://github.com/home-assistant-libs/pychromecast/blob/7372f8d24209aad9b5f7f61f5988f28994a7ec60/pychromecast/controllers/__init__.py#L76-L81

Eerovil: Huh, that's right. I guess that should also make the new check in `quick_play.quick_play` redundant.

I did get a strange error telling me something like "chromecast is not ready" previously... I'll need to check that again.

Eerovil:
Everything seems to work now with the latest changes from https://github.com/home-assistant/core/pull/35114

Also, I've successfully converted [Cast Extensions](https://github.com/Eerovil/castextensions) into a separate custom component.

Eerovil:
One question to think about, is whether we should have all controllers implement the `quick_play()` function, or just implement all quick_play functionality in `quick_play.quick_play`? I'm starting to think the latter could actually be cleaner, since there will be controller-specific code anyway.

emontnemery:
I think it's better to keep controller specific stuff in the controllers.

Actually, why is this needed in `quick_play.quick_play()`:
        kwargs = {
            "video_id": data.pop("media_id"),
            "enqueue": data.pop("enqueue", False),
            "playlist_id": data.pop("playlist_id", None),
        }
Isn't it simpler to change the prototype of def `quick_play(cast, app_name, data):` to `def quick_play(cast, app_name, **kwargs):` and let the controller pick out the keys it needs?

Eerovil:
Yes, that would be better! The current way made more sense when all the keys were predefined, but now we can move all that into the controller class.

emontnemery:
@Eerovil Are you happy with this PR now?

Eerovil:
Yes, It's ready.

PR #362, opened 2020-05-13
Add ability to pass in a zeroconf instance to discovery

bdraco:
This allows updating `cast` to use the shared zeroconf instance added in home-assistant/core#35484


bdraco:
Realized this probably needs to do `browser.cancel()` to stop the browser as `.close()` is protected in Home Assistant right after submitting.

bdraco:
Verified google cast was still discovered and tested with `cast.show_lovelace_view`

PR #363, opened 2020-05-25
Ignore faulty data, making "not iterable" error

Ebbe:
I live in Denmark, and here the public station "DR" have a chromecast app. When watching a series, and the "Next episode" teaser appears on the TV, python throws an exception.

```
Exception in thread Thread-1:
  File "/usr/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/home/ebbe/code/ebbe/tmp/pychromecast/pychromecast/socket_client.py", line 531, in run
    if self.run_once(timeout=POLL_TIME_BLOCKING) == 1:
  File "/home/ebbe/code/ebbe/tmp/pychromecast/pychromecast/socket_client.py", line 607, in run_once
    if REQUEST_ID in data:
TypeError: argument of type 'int' is not iterable
```

I debugged `data`, and found it sometimes (near the end of an episode) contained integers: 1001 and 1002. This `try` fixes the issue.

Discovered the error through Home Assistant, when it simply stopped being updated (because the python thread had failed).

### DIFF
--- a/pychromecast/socket_client.py
+++ b/pychromecast/socket_client.py
@@ -604,15 +604,18 @@ class SocketClient(threading.Thread):
         # See if any handlers will accept this message
         self._route_message(message, data)
 
-        if REQUEST_ID in data:
-            callback = self._request_callbacks.pop(data[REQUEST_ID], None)
-            if callback is not None:
-                event = callback["event"]
-                callback["response"] = data
-                function = callback["function"]
-                event.set()
-                if function:
-                    function(data)
+        try:
+          if REQUEST_ID in data:
+              callback = self._request_callbacks.pop(data[REQUEST_ID], None)
+              if callback is not None:
+                  event = callback["event"]
+                  callback["response"] = data
+                  function = callback["function"]
+                  event.set()
+                  if function:
+                      function(data)
+        except:
+          pass # Invalid data. Ignoring
 
         return 0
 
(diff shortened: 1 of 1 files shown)


balloob:
You can't just throw a guard around it… that's ignoring the error, not fixing the problem. 

The old Cast protocol was based on a websocket, so I wouldn't be surprised if these are websocket status codes

```

   1001

      1001 indicates that an endpoint is "going away", such as a server
      going down or a browser having navigated away from a page.

   1002

      1002 indicates that an endpoint is terminating the connection due
      to a protocol error.
```

(from https://tools.ietf.org/html/rfc6455#section-7.4.1)

Ebbe:
I tried detecting if it's an int. If that's not what acceptable, I need a pointer to the right direction. At least it doesn't crash for me any more.

Ebbe:
@balloob Is there still an issue with my pr? I'll gladly refactor it if you point me in the right direction.

emontnemery:
@Ebbe Sorry for the lack of feedback!
Can you verify this PR https://github.com/home-assistant-libs/pychromecast/pull/405 solves your problem?

Ebbe:
> @Ebbe Sorry for the lack of feedback!
> Can you verify this PR #405 solves your problem?

Yes! I just tried it, and it works :) Thanks! Closing my pull request.

PR #364, opened 2020-05-25
Handle zeroconf update_service

emontnemery:


pychromecast/discovery.py
```diff
@@ -42,6 +42,11 @@ def remove_service(self, zconf, typ, name):
         if self.remove_callback:
             self.remove_callback(name, service)
 
+    def update_service(self, zconf, typ, name):
+        """ Update a service in the collection. """
+        _LOGGER.debug("update_service %s, %s", typ, name)
+        self.add_service(zconf, typ, name)
```
balloob, line 48: Now this will also log `add_service`. Maybe extract the handling and have both `update` and `add` call it ?

emontnemery: Fixed!

PR #370, opened 2020-06-01
Refactor start_discovery to fix race

emontnemery:
Prevent a race in `start_discovery` where callbacks arrive before the listener is returned.

Fix for home-assistant/core#36326

### DIFF
--- a/pychromecast/discovery.py
+++ b/pychromecast/discovery.py
@@ -94,12 +94,7 @@ class CastListener:
             callback(name)
 
 
-def start_discovery(
-    add_callback=None,
-    remove_callback=None,
-    update_callback=None,
-    zeroconf_instance=None,
-):
+def start_discovery(listener, zeroconf_instance=None):
     """
     Start discovering chromecasts on the network.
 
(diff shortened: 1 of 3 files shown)


pychromecast/discovery.py
```diff
@@ -108,28 +103,31 @@ def start_discovery(
...
 
 
 def stop_discovery(browser):
     """Stop the chromecast discovery thread."""
-    browser.cancel()
+    try:
+        browser.cancel()
+    except RuntimeError:
+        # Throws if called from service callback when joining the zc browser thread
+        pass
```
emontnemery, line 130: @bdraco `browser.cancel()` was added in #362. It throws if called from a service callback when trying to join. Not sure if the error should be ignored here, or if zeroconf should handle it?

bdraco: Since a thread can't wait for itself and `self.done` gets set to `True` trapping exception makes sense here.

zeroconf should probably be updated to skip the `.join` when `threading.current_thread()` is the `ServiceBrowser` thread

balloob: We can use Thread Local data to check if we're in service browser data: https://docs.python.org/3.7/library/threading.html#thread-local-data

Issue #381, opened 2020-06-30
Chromecast object has None host


tareco:
Hi,

I am getting always the value "None" for the host attribute of any Chromecast:

```py
import pychromecast
chromecasts = pychromecast.get_chromecasts()
for cc in chromecasts[0]:
  cc.wait()
  print(cc)
```

`Chromecast(None, port=8009, device=DeviceStatus(friendly_name...`

I am not sure if this is a bug or if I'm doing something wrong, but I can get lots of info from the CC including the current active App etc.. only the "host" never gives me an IP address.

Thank you.

emontnemery:
It's not a bug, `pychromecast.get_chromecasts()` gets a list of chromecasts by using mDNS discovery, and will create Chromecast objects based on mDNS service data, not based on IP or host name.

Using IP or host name is strongly advised against as IP or host name may change over time, especially for audio groups where the group leader changes over time.

tareco:
Ok, I understand. So this is now off-topic.. but how can I get the current IP address of the Chromecast? I am exploring the feasibility of configuring remotely some settings, such as the wifi network, and to POST data I'll need the IP. Thank you!

emontnemery:
Sure, you can do like this:
https://github.com/home-assistant/core/blob/38210ebbc6a06220b3be3a29259e13683b693f4e/homeassistant/components/cast/discovery.py#L57-L81

```py
        # For support of deprecated IP based white listing
        zconf = ChromeCastZeroconf.get_zeroconf()
        service_info = None
        tries = 0
        while service_info is None and tries < 4:
            try:
                service_info = zconf.get_service_info(
                    "_googlecast._tcp.local.", service_name
                )
            except OSError:
                # If the zeroconf fails to receive the necessary data we abort
                # adding the service
                break
            tries += 1


        if not service_info:
            _LOGGER.warning(
                "setup_internal_discovery failed to get info for %s, %s",
                uuid,
                service_name,
            )
            return


        addresses = service_info.parsed_addresses()
        host = addresses[0] if addresses else service_info.server
```

theychx:
@emontnemery 
> you can get the service_name from the chromecast object

Forgive my ignorance, but I can't seem to find it (the service_name).

PS:
Isn't [this](https://github.com/home-assistant-libs/pychromecast/blob/f3f8449d6edb3ec340f7172bd948cca20d52c697/pychromecast/__init__.py#L268) a mistake? Shouldn't it default to `None`?

[...]

PR #383, opened 2020-06-30
include host for returned devices

kalkun:
The recent release of version 7 did not include host name for the discovered devices. I have tried to reintroduce this feature with this PR in a similar way to what existed before this release.

### DIFF
--- a/pychromecast/__init__.py
+++ b/pychromecast/__init__.py
@@ -70,7 +70,7 @@ def get_chromecast_from_service(
     # Build device status from the mDNS service name info, this
     # information is the primary source and the remaining will be
     # fetched later on.
-    services, uuid, model_name, friendly_name = services
+    host, services, uuid, model_name, friendly_name = services
     _LOGGER.debug("_get_chromecast_from_service %s", services)
     cast_type = CAST_TYPES.get(model_name.lower(), CAST_TYPE_CHROMECAST)
     manufacturer = CAST_MANUFACTURERS.get(model_name.lower(), "Google Inc.")
@@ -82,7 +82,7 @@ def get_chromecast_from_service(
         cast_type=cast_type,
     )
     return Chromecast(
-        host=None,
+        host=host,
         device=device,
         tries=tries,
         timeout=timeout,
--- a/pychromecast/discovery.py
+++ b/pychromecast/discovery.py
@@ -99,11 +99,16 @@ class CastListener:
             return
         uuid = UUID(uuid)
 
+        hosts = service.parsed_addresses()
+        if len(hosts):
+            host = hosts[0]
+
         services_for_uuid = self.services.setdefault(
             uuid, ({name}, uuid, model_name, friendly_name)
         )
         services_for_uuid[0].add(name)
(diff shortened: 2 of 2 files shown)


emontnemery:
The change is intentional, including `host` in the returned discovery data is very misleading because there is no guarantee the `host` is stable. This is particularly problematic for audio groups where group leadership, and hence host, is passed around among the members.

Note that the `socket_client` supports resolving a set of service names to IP/hsot, and will do so each time it's (re)connecting.

Why is it useful to have the `host` set in the discovered devices?

kalkun:
alright, thank  you! I might have jumped the gun a little. I did see the breaking changes disclaimer but not the reason why. I have some automation over another tool that i do not have control over that is then fed to a cvlc command to start casting based on the host found via `pychromecast`.

skorokithakis:
We use the host in `catt` to find the device to connect to, so this change has broken discovery. What is the recommended way of handling this now?

emontnemery:
OK, can you use this as a workaround: https://github.com/home-assistant-libs/pychromecast/issues/381 for now?

Also, can you explain a bit more about your use case?
Do you use pychromecast only for discovery, you don't use pychromecast's `socket_client` to talk to the cast device?

kalkun:
Yea, i only needed the host in order to build a cvlc command that i can pipe to from another program.

skorokithakis:
> OK, can you use this as a workaround: [#381 (comment)](https://github.com/home-assistant-libs/pychromecast/issues/381#issuecomment-651945670) for now?

Yep, that should work, thank you.

> Also, can you explain a bit more about your use case?
> Do you use pychromecast only for discovery, you don't use pychromecast's `socket_client` to talk to the cast device?

We use PyChromecast to discover the devices and then use the IP to talk to them, we don't keep the socket. We also cache the IP to make connection quicker next time (ie so we can skip discovery if the cached IP works).

emontnemery:
OK, I see.

Instead of `get_chromecasts` you can use `get_listed_chromecasts` instead, and pass it a lists of the wanted names, it will be almost instant, although it's still a bit heavyweight.
Please note you need to close the browser which is now returned by the `get_chromecasts` function and variants if you don't need mDNS updates.

For what you are doing, please have a look at `examples/discovery_examples.py` it does only the discovery.

The changes in this PR to `pychromecast/discovery.py` are OK, maybe with some helpers similar to `get_chromecasts`, `get_listed_chromecasts` etc: `discover_chromecasts`, `discover_listed_chromecasts` which does the discovery only.

kalkun:
Alright, I will try that  thank you!

balloob:
If you just need to discovery chromecasts, no need to use pychromecast at all. Just rely on zeroconf and search for the `"_googlecast._tcp.local."` type.

skorokithakis:
@balloob That's actually a very good idea, thank you. I will see if we can transition to that instead.

emontnemery:
I opened #385 which includes the changes in this PR to `pychromecast/discovery.py` as well as some new helpers and examples.
@kalkun I think this PR can be closed in favor of #385?

PR #385, opened 2020-07-01
Report host+port for discovered casts

emontnemery:
Report host+port for discovered casts
Add new discovery helper `discover_listed_chromecasts`
Add additional discovery examples

Fix default value of zconf in Chromecast constructor

### DIFF
--- a/examples/discovery_example2.py
+++ b/examples/discovery_example2.py
@@ -1,0 +1,25 @@
+"""
+Example that shows how to list available chromecasts.
+"""
+import argparse
+import logging
+import time
+
+import pychromecast
+import zeroconf
+
+parser = argparse.ArgumentParser(description="Example on how to receive updates on discovered chromecasts.")
+parser.add_argument("--show-debug", help="Enable debug log", action="store_true")
+args = parser.parse_args()
+
+if args.show_debug:
+    logging.basicConfig(level=logging.DEBUG)
+
+devices, browser = pychromecast.discovery.discover_chromecasts()
+# Shut down discovery
+pychromecast.stop_discovery(browser)
+
+print(f"Discovered {len(devices)} device(s):")
+for device in devices:
+    print(f"  {device}")
+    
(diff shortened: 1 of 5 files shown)


examples/discovery_example2.py
```diff
@@ -0,0 +1,25 @@
...
+parser = argparse.ArgumentParser(description="Example on how to receive updates on discovered chromecasts.")
+parser.add_argument("--show-debug", help="Enable debug log", action="store_true")
+args = parser.parse_args()
+
+if args.show_debug:
+    logging.basicConfig(level=logging.DEBUG)
+
+devices, browser = pychromecast.discovery.discover_chromecasts()
+# Shut down discovery
+pychromecast.stop_discovery(browser)
```
tareco, line 20: I am getting this:
AttributeError: module 'pychromecast' has no attribute 'stop_discovery'

emontnemery: Did you include all the changes from the PR?
`stop_discovery` is readded to pychromecast here: https://github.com/home-assistant-libs/pychromecast/blob/b43127e0e6d884356ef3622d72a468e2eb544346/pychromecast/__init__.py#L21

pychromecast/discovery.py
```diff
@@ -166,6 +171,47 @@ def callback(uuid, name):
     return (listener.devices, browser)
 
 
+def discover_listed_chromecasts(
```
balloob, line 174: One more discovery method 😆 Should we remove some other ones?

examples/discovery_example2.py
tareco: oops.. you're right :P

pychromecast/discovery.py
emontnemery: I think it's fairly OK:
 - `discover_chromecasts` and `discover_listed_chromecasts` are helpers for discovery only.
 - `get_chromecasts` and `get_listed_chromecasts` are discovers and creates Chromecast objects.

PR #391, opened 2020-07-04
update_status of MediaController after connect

dominikkarall:
I think it makes sense to update the media status after connection, otherwise listeners from users wouldn't be called and old values may remain.
Please test this change before merge as I can't test with my devices at the moment.

### DIFF
--- a/pychromecast/socket_client.py
+++ b/pychromecast/socket_client.py
@@ -358,6 +358,7 @@ class SocketClient(threading.Thread):
                         )
                     )
                     self.receiver_controller.update_status()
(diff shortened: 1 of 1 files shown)


dominikkarall:
I just saw that the update is called in `channel_connected` function in `media.py`. Therefore this change is useless.

emontnemery:
Did you have some problem with media status being stale after reconnect though, or it was just a misunderstanding?

dominikkarall:
I had an issue that the media status wasn't updated on first connection. If it happens again, I will investigate in more detail.

Issue #375, opened 2020-06-21
cast.reboot() does not work


surlyhacker:
I tried a variant of the sample code from the README to see if I could reboot a Google Home Mini.  Nothing happens when I call reboot() even after waiting several minutes.  It still responds to volume commands immediately after calling reboot() too so I am pretty sure it's not rebooting.  (it normally has a light and sound pattern that clearly indicates a reboot)

```
>>> import time
>>> import pychromecast
>>> chromecasts = pychromecast.get_chromecasts()
>>> [cc.device.friendly_name for cc in chromecasts]
['one', 'two', 'three', 'four', 'five', 'six', 'seven', 'eight']
>>> cast = next(cc for cc in chromecasts if cc.device.friendly_name == "one")
>>> cast.wait()
>>> print(cast.device)
DeviceStatus(friendly_name='one', model_name='Google Home Mini', manufacturer='Google Inc.', uuid=UUID('redacted'), cast_type='audio')
>>> cast.reboot()
>>> cast.volume_up()
0.5099999964237213
>>> cast.volume_down()
0.40999999046325686
```

balloob:
Wouldn't be surprised if that API has been removed. It's an older one.

surlyhacker:
Well elsewhere I've seen research that shows that it seems to need an auth token now, and tried it a bit myself:

See this issue/comment: https://github.com/rithvikvibhu/GHLocalApi/issues/39#issuecomment-644959103

```
# curl -Lv --insecure -H Content-Type:application/json --data-raw '{"params":"now"}' https://redacted:8443/setup/reboot
...
> POST /setup/reboot HTTP/1.1
...
< HTTP/1.1 401 Unauthorized
```

But shouldn't pychromecast throw an error if it receives a 401?

dominikkarall:
It seems that reboot doesn't work at all with the new get_chromecast_from_service usage. Host is always None and therefore it can't connect to the device.

balloob:
@emontnemery we should remove that function. With the recent cast changes we are unable to reboot anyway.

emontnemery:
Yeah, I'll remove it.

PR #394, opened 2020-07-08
Remove reboot function

emontnemery:
Remove `reboot()` function, it doesn't work any more after Google locked down the local API.
See #375 and https://github.com/rithvikvibhu/GHLocalApi/issues/39#issuecomment-644959103

### DIFF
--- a/pychromecast/__init__.py
+++ b/pychromecast/__init__.py
@@ -440,10 +440,6 @@ class Chromecast:
 
         self.socket_client.receiver_controller.stop_app()
 
-    def reboot(self):
-        """ Reboots the Chromecast. """
-        reboot(self.host)
-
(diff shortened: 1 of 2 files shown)


pychromecast/__init__.py
```diff
@@ -20,7 +20,7 @@
     start_discovery,
     stop_discovery,
 )
-from .dial import get_device_status, reboot, DeviceStatus
+from .dial import get_device_status, DeviceStatus
```
balloob, line 23: Does `get_device_status` still work ?

emontnemery: Yes, it still works, but it's crippled and doesn't reveal anything not revealed through mDNS discovery.
It could be removed, but then mDNS would be mandatory.

PR #396, opened 2020-07-08
support non-blocking reconnect

dominikkarall:
This change allows the usage of run_once() to reconnect the cast device in a non-blocking way. Every run_once() call it just tries to reconnect the device once for timeout period and finishes the function afterwards.

### DIFF
--- a/examples/asyncio_loop.py
+++ b/examples/asyncio_loop.py
@@ -1,0 +1,113 @@
+"""
+Example that shows how the socket client can be used without its own worker
+thread by not calling Chromecast.start() or Chromecast.wait(), but instead calling
+Chromecast.connect() and use asyncio as main loop.
+"""
+import argparse
+import logging
+import select
+import time
+import asyncio
+
+import pychromecast
+
+CAST_NAME = "Living Room"
+
+"""
+Check for cast.socket_client.get_socket() and
+handle it with cast.socket_client.run_once()
+"""
+
+loop = asyncio.get_event_loop()
+
+async def startChromecast():
+
+    cc = None
+    socketInLoop = 0
+    checkConnectionTask = None
+
+    def ccRunOnce():
+        cc.socket_client.run_once()
+
+    async def checkConnection():
+        nonlocal socketInLoop
+        while True:
+            try:
+                ccRunOnce()
+                if socketInLoop != cc.socket_client.get_socket() and cc.socket_client.is_connected:
+                    # update socket after reconnect
+                    socketInLoop = cc.socket_client.get_socket()
+                    loop.add_reader(socketInLoop, ccRunOnce)
+                socketInLoop = cc.socket_client.get_socket()
+                loop.add_reader(socketInLoop, ccRunOnce)
+            except Exception as e:
+                print("ERROR: run_once: " + str(e))
+            await asyncio.sleep(2)
+
+    async def connectChromecast():
+        nonlocal checkConnectionTask
+        cc.connect()
+        checkConnectionTask = loop.create_task(checkConnection())
+
+    def castFound(chromecast):
+        if chromecast.name == CAST_NAME:
+            print("=> Discovered cast " + CAST_NAME)
+            nonlocal cc
+            cc = chromecast
+            loop.create_task(connectChromecast())
+
+    browser = pychromecast.get_chromecasts(blocking=False, tries=1, retry_wait=0.1, timeout=0.1, callback=castFound)
+
+    while True:
+        if cc is not None and cc.socket_client.is_connected:
+            print("=> Start BigBuckBunny")
+            cc.play_media(
+                (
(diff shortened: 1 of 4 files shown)


emontnemery:
I don't get it, why not just set `tries` to 1?

dominikkarall:
Unfortunately running `socket_client` in an own thread isn't always the best solution. In one of my projects with pychromecast I'm using `Inline::Python` to call Python from Perl which doesn't support a thread to run, it stops as soon as the function was processed. Furthermore running `socket_client` in an own thread requires proper handling of callbacks. Concurrent access of variables from main loop and another thread might make problems.
So actually the question might also be, do we really need a separate thread if there is already a main loop when using pychromecast as a library? Currently I'm running pychromecast together with `asyncio` and therefore I want `asyncio` to be the only main loop.

I understand your point that it makes more sense to rewrite/split `initialize_connection` instead of introducing the `blocking` stuff. That's definitely the better solution. I'll have a look into it and update the PR afterwards. Furthermore I'll also add an `asyncio` example code as I think that that might help others as well.

dominikkarall:
I split the `initialize_connection` function up into `initialize_connection_once` and `initialize_connection`. Nevertheless I need the information about `blocking` in the SocketClient, therefore the parameter needs to remain.
In `_connection_check` I determine if it is blocking or not and if it's not I just call `initialize_connection_once` instead of `initialize_connection`.

I added a `asyncio` example as well, which shows the non-blocking functionality in combination with `asyncio`.

Please let me know if that change is acceptable for you, if not, let me know which parts should be adapted. I just want to have a perfect non-blocking pychromecast working for everyone.

Btw, can you please help me with the failed black test. I don't know what to do based on the output of black. Thx!

emontnemery:
With the current change, behavior will be quite different if blocking is set compared to when it's not. That will make the code more difficult to maintain.

Can you try to move the counting of retries etc such that it's done in context of `run_once`, which can then return to `run` after each retry?
In that way, there would be a single behavior and there is no need have a `blocking` parameter.

dominikkarall:
I will try to move retries to `run`, that's the best place as `run` is only called by blocking thread. In that case, `run_once` would always be a completely non-blocking function (as long as timeout is low) and can be called directly from non-blocking code. That  would definitely make it easier to maintain.

dominikkarall:
I moved tries routine to `run` and also added most of the exceptions to `run` which stop pychromecast when run in blocking mode. `retries` per service should be properly handled again, as that wasn't the case on my first commit.

I rewrote also the `first_connection` part, as this one was ahead of `run_once` which actually isn't necessary. Now it's part of `_check_connection` and the only difference is, that it sets the status to `disconnected` instead of `lost`.

dominikkarall:
I just did a few more tests and found some bugs. I'll update the PR as soon as I fixed them.

dominikkarall:
Ok, it works now :)
I found one more function (`get_info_from_service`) which had a timeout of 3s in discovery. I reduced it to 50ms to have some sort of "non-blocking".

dominikkarall:
Have you had a chance to review the PR? I just want to let you know that it runs without any issues in my asyncio application. Re-connection (also for nightly reboots) works perfectly.

emontnemery:
Sorry, I didn't have time to review+test yet.

By the way, even with this change the `connect()` calls may block for quite some time (seconds) if the cast is down.
Is that OK for your use case, i.e. your problem was that `run_once` would block forever if the cast was down?

dominikkarall:
Thanks for the hint. I think that should be handled by `socket.setTimeout()`.
https://github.com/home-assistant-libs/pychromecast/blob/master/pychromecast/socket_client.py#L293

Actually it didn't block when the device was down, but I couldn't reconnect with `run_once` as `tries` was always 1. I had to initiate a completely new discovery process and handle the result. With this change `run_once` just needs to be called again in the main loop and it automatically reconnects.

dominikkarall:
I would recommend to close this PR as I think the change might be to big.

Nevertheless, what do you think about making pychromecast asyncio ready by actually using asyncio? That would make the separate thread obsolete as well.

emontnemery:
Sorry for not responding!
I agree that the change was big and it didn't fully accomplish goal since there were still blocking calls in the code (although with timeouts).

Rewriting to use asyncio for the communication is interesting, but it would be a huge breaking change.

Note that the library is asyncio friendly, in the sense that nothing blocks, if you let the socketclient have its own thread and you don't use the blocking discovery functions but rely on callbacks instead. This is HomeAssistant's pychromecast wrapper works for example.

dominikkarall:
If I have some more spare time, I'll give it a try to implement asyncio. I think a lot of new stuff will rely on it.

I found one blocking issue in the YouTube controller, but it would be possible to do some smaller fixes there to allow also non blocking usage if asyncio is used on wrapper side. I'll open a PR for that the next days.

PR #397, opened 2020-07-09
Fixed error with Youtube exemple

Tagge:
Previous version caused Python to show a "Object is non-subscriptable" error. Converting it to a list fixes the problem.


emontnemery:
Thanks for the PR! This has been fixed by changing the return type to a list in #400 though.

PR #400, opened 2020-07-29
Allow passing an already existing zeroconf object to additional functions

emontnemery:
Allow passing an already existing zeroconf object to additional functions
Ignore browser updates for *._sub._googlecast._tcp.local.
Fix return type of `get_listed_chromecasts`

### DIFF
--- a/pychromecast/discovery.py
+++ b/pychromecast/discovery.py
@@ -168,7 +168,7 @@ def discover_chromecasts(
       A service browser to keep the Chromecast mDNS data updated. When updates
       are (no longer) needed, pass the browser object to
       pychromecast.discovery.stop_discovery().
-      
+
     :param zeroconf_instance: An existing zeroconf instance.
     """
     # pylint: disable=unused-argument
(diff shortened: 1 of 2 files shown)


pychromecast/discovery.py
```diff
@@ -190,13 +202,14 @@ def discover_listed_chromecasts(
       A list of chromecast services matching the criteria,
       or an empty list if no matching chromecasts were found.
       A service browser to keep the Chromecast mDNS data updated. When updates
-      are (no longer) needed, pass the broswer object to
-      pychromecast.discovery.stop_discover().
+      are (no longer) needed, pass the browser object to
+      pychromecast.discovery.stop_discovery().
```
balloob, line 206: Will stop discovery stop the zeroconf instance that the browser is connected to ? It should do so only if the browser is not reusing a zeroconf instance?

emontnemery: HomeAssistant's shared zeroconf instance overrides the `close()` method so it's not a problem.

PR #404, opened 2020-08-16
add support for QUEUE_INSERT

bjesus:
This is aiming to solve issue https://github.com/home-assistant-libs/pychromecast/issues/403 by allowing users to queue new items after the currently playing track.

I've added a `queue_media()` method that works just like `play_media()`, but instead of playing the media immediately it queues it. The Chromecast will automatically play it after the currently playing (and currently queued) media, and users can also skip to it using `queue_next()` etc.

I've also added an example.


bjesus:
I've fixed some formatting issues and rebased.

emontnemery:
The new code is almost exactly the same as `play_media` + its helper functions.
Can you try refactoring to avoid duplicating so much?

bjesus:
Sure, I can try. Should there still be a `queue_media()` method, or should I just add an optional argument to `play_media()` (eg. `queue=True`) that makes the item queued instead of played?

emontnemery:
An optional parameter `enqueue` sounds good!

bjesus:
I don't have any changes planned :) Thank you!

bjesus:
Is there anything I should fix to help this getting merged?

emontnemery:
Thanks! Do you need a release with the QUEUE_INSERT functionality?

bjesus:
That would be great as I could then submit a patch to `catt` to support this functionality! Thank you!

emontnemery:
OK, it's included in release 7.3.0

PR #429, opened 2020-11-15
Spotify quick play

Eerovil:
I've copied the spotify example code to allow easily casting spotify URI:s from home assistant.

This one requires two extra (small) dependencies. I'm not sure if importing them like this is smart?


Eerovil:
Apparently spotipy requires beautifulsoup which is pretty big.

balloob:
We don't want to add this to PyChromecast. Consumers of PyChromecast need to get their own token and pass it into the lib. This lib will not contain logic to get auth tokens for various services, just control their apps.

PR #430, opened 2020-11-15
Spotify quick play (again)

Eerovil:
I'm putting this up without any actual playing functionality. Since homeassistant has an interface through quick_play, we'll use that to launch the spotify app.

Together with this, homeassitant and a homeassistant custom component it's possible to cast any spotify URI (First get the access token, then use quick_play, then use spotipy to play the URI).

Just to note, I've moved the offending code here: https://github.com/Eerovil/castextensions/blob/master/custom_components/cast_extensions/app_spotify.py


pychromecast/controllers/spotify.py
```diff
@@ -81,10 +82,25 @@ def callback():
...
             )
+
+    # pylint: disable=too-many-locals
+    def quick_play(self, **kwargs):
+        """
+        Launches the spotify controller and returns when it's ready.
+        To actually play media, another application using spotify connect is required.
+        """
+        self.access_token = kwargs["access_token"]
+        self.expires = kwargs["expires"]
```
balloob, line 104:
```suggestion
    def quick_play(self, access_token, expires):
        """
        Launches the spotify controller and returns when it's ready.
        To actually play media, another application using spotify connect is required.
        """
        self.access_token = access_token
        self.expires = expires
```

Eerovil: All the other quick_play functions use purely kwargs, so I can't really switch to positional here. I can, though use `access_token=None, expires=None` and a check afterwards.

PR #437, opened 2020-12-19
Update local http API, re-add get_multizone_status

emontnemery:


pychromecast/dial.py
```diff
@@ -35,7 +38,13 @@ def _get_status(host, services, zconf, path):
 
     headers = {"content-type": "application/json"}
 
-    req = requests.get(FORMAT_BASE_URL.format(host) + path, headers=headers, timeout=10)
+    if secure:
+        url = FORMAT_BASE_URL_HTTPS.format(host) + path
+    else:
+        url = FORMAT_BASE_URL_HTTP.format(host) + path
+    _LOGGER.error("url: %s", url)
+    urllib3.disable_warnings(category=InsecureRequestWarning)
```
emontnemery, line 46: The casts don't have valid certificates, is this an OK way to disable warnings, or will it affect other libraries using requests?

PR #449, opened 2021-01-23
Add set_playback_rate to MediaController

DorianRudolph:


emontnemery:
This looks pretty good!
Can you add an example script which uses the functionality?

emontnemery:
I'm closing this since there has been no response in more than a year.

The change is welcome though, but there should be an example of how this can be used.
Please reopen or create a new PR.

PR #457, opened 2021-01-29
Infer contentType from playQueue items

maykar:
PlexAPI now has a getPlayQueue method that grabs the playQueue from server. This is now used in a [new service to play a playQueue](https://github.com/home-assistant/core/pull/45580) on plex capable media_players in HA 2021.2.0b1 (google cast devices, sonos, and plex clients). This fixes that service failing with:

`AttributeError: 'PlayQueue' object has no attribute 'playQueueType'`

playQueues from server have no playQueueType only PlexAPI playQueues have that attribute.


jjlawren (approved):
This is how it's handled in `plexapi` for native Plex clients. Looks good 👍

PR #469, opened 2021-02-25
Update discovery to support a list of known hosts

emontnemery:
### Breaking change:
Discovery functionality has been changed

### Suggested change:
Update discovery to support a list of known hosts

### DIFF
--- a/pychromecast/discovery.py
+++ b/pychromecast/discovery.py
@@ -162,14 +162,18 @@ class ZeroConfListener:
             return
         uuid = UUID(uuid)
 
-        cast_info = self._devices.setdefault(
-            uuid, CastInfo(set(), uuid, model_name, friendly_name, host, service.port)
-        )
-        # Update stored information
-        cast_info.services.add(ServiceInfo(SERVICE_TYPE_MDNS, name))
-        self._devices[uuid] = CastInfo(
-            cast_info.services, uuid, model_name, friendly_name, host, service.port
-        )
+        service_info = ServiceInfo(SERVICE_TYPE_MDNS, name)
+        if uuid not in self._devices:
+            self._devices[uuid] = CastInfo(
+                {service_info}, uuid, model_name, friendly_name, host, service.port
+            )
+        else:
+            # Update stored information
+            services = self._devices[uuid].services
+            services.add(service_info)
+            self._devices[uuid] = CastInfo(
+                services, uuid, model_name, friendly_name, host, service.port
+            )
 
         callback(uuid, name)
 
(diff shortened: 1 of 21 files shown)


pychromecast/discovery.py
```diff
@@ -1,68 +1,120 @@
...
 
 import zeroconf
 
+from .const import SERVICE_TYPE_HOST, SERVICE_TYPE_MDNS
+from .dial import get_device_status, get_multizone_status
+
 DISCOVER_TIMEOUT = 5
 
+ServiceInfo = namedtuple("ServiceInfo", ["type", "data"])
+CastInfo = namedtuple(
```
balloob, line 18: Are we ready yet to migrate to dataclasses?

balloob: I guess the downside for `ServiceInfo` is that it won't be hashable unless it's frozen. Guess not so useful besides some typing help.

pychromecast/discovery.py
```diff
@@ -106,56 +162,252 @@ def get_value(key):
...
+        self._devices = devices
+        self._known_hosts = {}
+        self._next_update = time.time()
+        self._start_requested = False
+        self.stop = threading.Event()
+
+    def add_hosts(self, known_hosts):
+        """Add a list of known hosts to the set."""
+        for host in known_hosts:
+            self._known_hosts.setdefault(host, HostStatus())
```
balloob, line 204: This could potentially create and throw away a lot of `HostStatus` objects. Why not check if key `host` is already set?

balloob (approved):

pychromecast/discovery.py
emontnemery: Thanks, fixed!

pychromecast/discovery.py
alexdelorenzo: [You can construct `namedtuple`s with type annotations like you can with `dataclasses` by subclassing  `typing.NamedTuple`](https://docs.python.org/3/library/typing.html#typing.NamedTuple): 
```python3
from typing import NamedTuple, Optional

class ServiceInfo(NamedTuple):
  type: str
  data: Optional[str] = None

example = ServiceInfo('example')
```

emontnemery: @alexdelorenzo There are no type annotations in pychromecast yet, would there be any benefit of using this?

alexdelorenzo: @emontnemery I bring it up because of @balloob's comment about moving to dataclasses and typing. There's no real benefit other than typing for namedtuples.

emontnemery: OK, thanks!

PR #484, opened 2021-04-03
Decrease discovery overhead

bdraco:
Attempt at reducing the discovery overhead

Before
Screen Shot 2021-04-02 at 6 26 36 PM

After -- all the time creating the ssl context is gone:

Screen Shot 2021-04-02 at 6 41 53 PM

### DIFF
--- a/pychromecast/dial.py
+++ b/pychromecast/dial.py
@@ -2,13 +2,11 @@
 Implements the DIAL-protocol to communicate with the Chromecast
 """
 from collections import namedtuple
-import json
 import logging
 import socket
-import ssl
-import urllib.request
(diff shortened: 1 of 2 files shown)


emontnemery:
The reason for using urllib directly instead of requests is that the latter did not accept the certificates presented by some chromecasts.
Are we sure this won't be a problem?
Can't we keep the SSL session around without going back to requests?

emontnemery:
I attempted to reuse the SSL context in #485, does that work?

bdraco:
Closing in favor of https://github.com/home-assistant-libs/pychromecast/pull/485

PR #485, opened 2021-04-03
Reuse SSL context in host browser

emontnemery:
Reuse SSL context in host browser

This is intended to bring the same benefits as #484, but without moving back to using the requests library


bdraco:
👍 Will test shortly

bdraco (approved):
Tested and working, verified the ssl setup performance drag is gone 👍  Nice!

PR #489, opened 2021-04-12
get_chromcasts() copy device list before iteration

Shane-Wiseman-Bose:
resolves https://github.com/home-assistant-libs/pychromecast/issues/488 by making a copy of device list returned from discovery before iterating over list.

### DIFF
--- a/pychromecast/discovery.py
+++ b/pychromecast/discovery.py
@@ -523,7 +523,7 @@ def discover_chromecasts(
     # Wait for the timeout or the maximum number of devices
     discover_complete.wait(timeout)
 
-    return (browser.devices.values(), browser)
+    return (list(browser.devices.values()), browser)
 
 
 def discover_listed_chromecasts(
@@ -572,4 +572,4 @@ def discover_listed_chromecasts(
 
     # Wait for the timeout or found all wanted devices
     discover_complete.wait(discovery_timeout)
-    return (cc_list.values(), browser)
+    return (list(cc_list.values()), browser)


Shane-Wiseman-Bose:
There is another option of making the copy on the iterable return from  discover_chromecasts()  which can protect all calls. I figured to lessen the scope for ease of review and to reflect my depth of the repos. 

https://github.com/home-assistant-libs/pychromecast/blob/48a1d59a8b69f4fee8604def94bb01a2f6815a64/pychromecast/discovery.py#L526

Shane-Wiseman-Bose:
### Example of issue.
```
>>> a = {"a": 1, "b": 2}
>>> b = a.values()
>>> for i in b:
...     if "a" in a:
...             del(a["a"])
...
  File "<stdin>", line 1, in <module>
RuntimeError: dictionary changed size during iteration
```
### Example of Fix
```
>>> a = {"a": 1, "b": 2}
>>> b = a.values()
>>> for i in list(b):
...     if "a" in a:
...             del(a["a"])
...
>>>
```

emontnemery:
> There is another option of making the copy on the iterable return from discover_chromecasts() which can protect all calls

That's a better option, can you update the PR to do that instead?

Shane-Wiseman-Bose:
moved the list copy to discovery.py calls. A user can still access the mutable device object by referencing the browser attribute _browser.devices_

Shane-Wiseman-Bose:
@emontnemery is there anything else pending on this?

emontnemery:
@Shane-Wiseman-Bose nothing else, thanks for the PR! 👍

PR #500, opened 2021-05-07
Add force_launch option to BaseController.launch

dphoyes:
This option already exists in Chromecast.start_app(). This change just makes the option conveniently available via the controller instance.


emontnemery:
@dphoyes can you fix the black error please?

emontnemery (approved):
Thanks, @dphoyes 👍

PR #503, opened 2021-05-27
Add support for BBC Sounds and iPlayer apps

blawford:
## Changes:
- Adds new controllers for BBC Sounds and BBC iPlayer, associated config and examples.
- Fixes a race in quick_play to snsure the app is started before playing media

There is also a draft [PR into HomeAssistant core](https://github.com/home-assistant/core/pull/51148) that makes a small change to allow metadata to be passed through when casting via an app that I would like to be reviewed alongside this. 

First time doing anything in python, so please bear that in mind when reviewing! If this all looks good then I will add another PR for updating the homeassistant documentation with examples for these apps.


pychromecast/controllers/bbcsounds.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def __init__(self):
+        super().__init__(APP_NAMESPACE, APP_BBCSOUNDS)
+
+        self.logger = logging.getLogger(__name__)
+
+    def play_media(self, media_id, is_live=False, **kwargs):
+        """
+        Play BBC Sounds media
+        """
```
emontnemery, line 24:
```suggestion
        """Play BBC Sounds media."""
```

pychromecast/quick_play.py
```diff
@@ -5,9 +5,11 @@
 from .controllers.yleareena import YleAreenaController
 from .controllers.spotify import SpotifyController
 from .controllers.bubbleupnp import BubbleUPNPController
+from .controllers.bbciplayer import BbcIplayerController
+from .controllers.bbcsounds import BbcSoundsController
 
 
-def quick_play(cast, app_name, data):
+def quick_play(cast, app_name, data, **kwargs):
```
emontnemery, line 12: Why do we need both `**kwargs` and a `data` dict?
Can Home Assistant's quickplay service pass in the needed `kwargs`?

pychromecast/controllers/bbciplayer.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def __init__(self):
+        super().__init__(APP_NAMESPACE, APP_BBCIPLAYER)
+
+        self.logger = logging.getLogger(__name__)
+
+    def play_media(self, media_id, is_live=False, **kwargs):
+        """
+        Play BBC iPlayer media
+        """
```
emontnemery, line 24:
```suggestion
        """Play BBC iPlayer media."""
```

examples/bbciplayer_example.py
```diff
@@ -0,0 +1,51 @@
...
+from pychromecast.controllers.bbciplayer import BbcIplayerController
+
+# Change to the name of your Chromecast
+CAST_NAME = "Lounge Video"
+
+# Note: Media ID is NOT the 8 digit alpha-numeric in the URL
+# it can be found by right clicking the playing video on the web interface
+# e.g. https://www.bbc.co.uk/iplayer/episode/b09w7fd9/bitz-bob-series-1-1-castle-makeover shows: 
+# "2908kbps | dash (mf_cloudfront_dash_https)
+#  b09w70r2 | 960x540"
```
emontnemery, line 21: This is a great explanation, add it in pychromecast/controllers/bbciplayer.py as well

examples/bbciplayer_example.py
```diff
@@ -0,0 +1,51 @@
...
+
+import logging
+from time import sleep
+import sys
+
+import pychromecast
+from pychromecast.controllers.bbciplayer import BbcIplayerController
+
+# Change to the name of your Chromecast
+CAST_NAME = "Lounge Video"
```
emontnemery, line 15: Copy the argument handling used in the other examples to make the cast name, debug level etc. configurable from command line 👍

examples/bbcsounds_example.py
```diff
@@ -0,0 +1,47 @@
...
+
+import logging
+from time import sleep
+import sys
+
+import pychromecast
+from pychromecast.controllers.bbcsounds import BbcSoundsController
+
+# Change to the name of your Chromecast
+CAST_NAME = "Lounge Video"
```
emontnemery, line 15: Copy the argument handling used in the other examples to make the cast name, debug level etc. configurable from command line 👍

examples/bbcsounds_example.py
```diff
@@ -0,0 +1,47 @@
...
+
+# Media ID can be found in the URL
+# e.g. https://www.bbc.co.uk/sounds/live:bbc_radio_one
```
emontnemery, line 18: This is a great explanation, add it in pychromecast/controllers/bbcsounds.py as well

emontnemery:
Looks good overall! 👍

pychromecast/quick_play.py
blawford: Not sure on your first point about whether both are needed, I will take another look. 
I have tested end-to-end (casting with and without metadata when calling the service in a script) with the changes in this PR and a one line change in the [draft PR](https://github.com/home-assistant/core/pull/51148).

emontnemery:
I tried the examples and they aren't working well, the problem is that there is a lack of synchronization after launching an app causing the `play_media` call to be ignored.

In `controllers/media.MediaController.play_media" this is handled by passing a callback function to the `launch_app` call.

The examples can be fixed by doing like this:
```py
bbcsounds = BbcSoundsController()
cast.register_handler(bbcsounds)
def app_launched_callback():
    """Plays media after chromecast has switched to requested app."""
    bbcsounds.play_media(MEDIA_ID, False, **METADATA)
bbcsounds.launch(callback_function=app_launched_callback)
```

The `quick_play` needs to be updated in the same way, could you give that a go?
I'd suggest to use the quick_play functions in the examples to make the examples more similar to what Home Assistant will do.

blawford:
> I tried the examples and they aren't working well, the problem is that there is a lack of synchronization after launching an app causing the `play_media` call to be ignored if the app wasn't already started.
> 
> In `controllers.media.MediaController.play_media` this is handled by passing a callback function to the `launch_app` call.
> 
> The examples can be fixed in a similar manner by doing like this:
> 
> ```python
> bbcsounds = BbcSoundsController()
> cast.register_handler(bbcsounds)
> def app_launched_callback():
>     """Plays media after chromecast has switched to requested app."""
>     bbcsounds.play_media(MEDIA_ID, False, **METADATA)
> bbcsounds.launch(callback_function=app_launched_callback)
> ```
> 
> The `quick_play` needs to be updated in the same way, could you give that a go?
> I'd suggest to use the quick_play functions in the examples to make the examples more similar to what Home Assistant will do.

Just to be sure, are you suggesting to update the 



> I tried the examples and they aren't working well, the problem is that there is a lack of synchronization after launching an app causing the `play_media` call to be ignored if the app wasn't already started.
> 
> In `controllers.media.MediaController.play_media` this is handled by passing a callback function to the `launch_app` call.
> 
> The examples can be fixed in a similar manner by doing like this:
> 
> ```python
> bbcsounds = BbcSoundsController()
> cast.register_handler(bbcsounds)
> def app_launched_callback():
>     """Plays media after chromecast has switched to requested app."""
>     bbcsounds.play_media(MEDIA_ID, False, **METADATA)
> bbcsounds.launch(callback_function=app_launched_callback)
> ```
> 
> The `quick_play` needs to be updated in the same way, could you give that a go?
> I'd suggest to use the quick_play functions in the examples to make the examples more similar to what Home Assistant will do.

Struggling to get this working, so I will double check I am understanding... Do you mean changing bbciplayer

> I tried the examples and they aren't working well, the problem is that there is a lack of synchronization after launching an app causing the `play_media` call to be ignored if the app wasn't already started.
> 
> In `controllers.media.MediaController.play_media` this is handled by passing a callback function to the `launch_app` call.
> 
> The examples can be fixed in a similar manner by doing like this:
> 
> ```python
> bbcsounds = BbcSoundsController()
> cast.register_handler(bbcsounds)
> def app_launched_callback():
>     """Plays media after chromecast has switched to requested app."""
>     bbcsounds.play_media(MEDIA_ID, False, **METADATA)
> bbcsounds.launch(callback_function=app_launched_callback)
> ```
> 
> The `quick_play` needs to be updated in the same way, could you give that a go?
> I'd suggest to use the quick_play functions in the examples to make the examples more similar to what Home Assistant will do.

Struggling to get this working and there is a chance I have miss understood what you asked. Could you have a look at [this commit](https://github.com/blawford/pychromecast/commit/0ff3df45cf90177a56db51add0772df34ad09552) and let me know if I am on the right track and have any suggestions why it isn't working.

It is launching the app but not playing the content.

emontnemery:
You're on the right track, but the correction for quick play should be in quick_play.quick_play, and the example should call quick_play.quick_play, to make it more similar to Home Assistant.

blawford:
> You're on the right track, but the correction for quick play should be in `quick_play.quick_play`, and the examples should also call `quick_play.quick_play`, to make the behavior more similar to Home Assistant.

Pushed changes, please review.

emontnemery:
Thanks, this looks good!

The examples seem to work reliably for me now, is it the same for you?

blawford:
@emontnemery - anything else or can this be merged?

emontnemery:
Thanks, @blawford 👍

PR #509, opened 2021-06-13
Fix spotify controller and update example

shocklateboy92:
It seems the `setCredentials` message has changed to `addUser` at some point in the past, with a different payload.

We also discovered a new issue with playback (on our google homes and chromecasts at least) where launching the spotify app and adding the user/token to it doesn't cause it to appear in spotify's device list. This updates the example to make the same request the web player does to enable playback on the cast device.

### DIFF
--- a/examples/spotify_example.py
+++ b/examples/spotify_example.py
@@ -114,8 +114,9 @@ headers = {
     "Authorization": "Bearer " + access_token,
 }
 transferResponse = requests.post(
-    "https://guc-spclient.spotify.com/connect-state/v1/connect/transfer/from/noop/to/"
-    + sp.device,
+    "https://guc-spclient.spotify.com/connect-state/v1/connect/transfer/from/noop/to/{}".format(
+        sp.device
+    ),
     headers=headers,
 )
 
--- a/examples/spotify_example.py
+++ b/examples/spotify_example.py
@@ -12,6 +12,7 @@ import argparse
 import logging
 import time
 import sys
+import requests
 
 import zeroconf
 import spotify_token as st  # pylint: disable=import-error
@@ -37,8 +38,8 @@ parser.add_argument("--show-debug", help="Enable debug log", action="store_true"
 parser.add_argument(
     "--show-zeroconf-debug", help="Enable zeroconf debug log", action="store_true"
 )
-parser.add_argument("--user", help="Spotify username", required=True)
-parser.add_argument("--password", help="Spotify password", required=True)
+parser.add_argument("--sp-key", help="Spotify cookie", required=True)
+parser.add_argument("--sp-dc", help="Spotify cookie", required=True)
 parser.add_argument(
     "--uri",
     help='Spotify uri(s) (default: "%(default)s")',
@@ -74,7 +75,7 @@ cast.wait()
 spotify_device_id = None
 
 # Create a spotify token
-data = st.start_session(args.user, args.password)
+data = st.start_session(args.sp_dc, args.sp_key)
 access_token = data[0]
 expires = data[1] - int(time.time())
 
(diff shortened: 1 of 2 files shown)


emontnemery:
@fondberg could you take a look at this please?

emontnemery:
@shocklateboy92 , @CraigDropTables how has this been tested? The pychromecast spotify functionality is used in some other places, and since we change the required arguments for starting the app, those need to be updated too?

examples/spotify_example.py
```diff
@@ -96,25 +105,30 @@
...
-devices_available = client.devices()
-
-# Match active spotify devices with the spotify controller's device id
-for device in devices_available["devices"]:
-    if device["id"] == sp.device:
-        spotify_device_id = device["id"]
-        break
+# The chromecast device does not show up as part of the public API get devices
+# call until it starts playing. The only way to do so is to transfer playback
+# by call this endpoint that's not part of their public API.
```
fondberg, line 110: Hmmm. I think this is not correct.
It will should show up. I dont need any special http calls

pychromecast/controllers/spotify.py
```diff
@@ -11,10 +11,9 @@
 APP_NAMESPACE = "urn:x-cast:com.spotify.chromecast.secure.v1"
 TYPE_GET_INFO = "getInfo"
 TYPE_GET_INFO_RESPONSE = "getInfoResponse"
-TYPE_SET_CREDENTIALS = "setCredentials"
```
fondberg, line 14: Why is this removed?

pychromecast/controllers/spotify.py
```diff
@@ -11,10 +11,9 @@
...
-TYPE_SET_CREDENTIALS_ERROR = "setCredentialsError"
-TYPE_SET_CREDENTIALS_RESPONSE = "setCredentialsResponse"
-
+TYPE_ADD_USER = "addUser"
```
fondberg, line 14: Can you describe what this new api call does?

shocklateboy92:
Sorry for the late response.

We tested this on our Google Home (and Google Home Mini). We originally tried it with Home Assistant, and noticed it made the "ding" noise but then got stuck. Looking at the logs, it looked like our Google Home as returning `setCredentialsError`.

So we followed the instructions on this project's readme.md and used net-export to look at what the spotify web player sends. 
That's when we noticed it doesn't send `setCredentials` at all, it sends an `addUser` in its place.

pychromecast/controllers/spotify.py
shocklateboy92: Based on the net-export log, our web player wasn't sending this anymore. We tried it with the same auth token, and it still returned `setCredentialsError`. Doing the same `addUser` the web player was sending worked.

pychromecast/controllers/spotify.py
shocklateboy92: As far as we can tell it does the same thing that `setCredentials` used to do, with the payload in a slightly different format.

examples/spotify_example.py
shocklateboy92: Odd. How long did it take to show up? Was it a Chromecast V1/V2? or Google Home/Mini? 
Did you use the auth token fetched by feeding your cookies to `spotify_token`?

fondberg:
interesting. so will this fix the expiry issue after 60 mins?

examples/spotify_example.py
fondberg: I maintain spotcast so I think I represent most of the users with a variety of cast devies

examples/spotify_example.py
```diff
@@ -96,25 +105,30 @@
...
+headers = {
+    "Accept": "application/json",
+    "Content-Type": "application/json",
+    "Authorization": "Bearer " + access_token,
+}
+transferResponse = requests.post(
+    "https://guc-spclient.spotify.com/connect-state/v1/connect/transfer/from/noop/to/{}".format(
```
fondberg, line 117: guc is for US.

It is not advisable to use spotify internal API:s. A simple force_playback with spotipy should suffice

rako77:
Clearly I did not test the existing functionality enough. I just tested Spotcast and it works for us so we may just need to update the example for now.

As Spotify is now using addUser instead of set credentials the setCredentials flow may break in the near future but I don't see a need to switch from setCredentials to addUser until the workflow breaks.

fondberg:
Seperate from this but... what _would_ be interesting is to see what happens after 60 mins when sniffing. Is there a new token sent?

emontnemery:
What's the way forwards for this PR?

fondberg:
@emontnemery seems they were right. The old setCredentials is not working anymore. 
@rako77 can you merge master and take this forward?

shocklateboy92:
Unfortunately @rako77 and I are a little squeezed for time at the moment.
We'll see what we can do when things calm down in a few weeks.

fondberg:
The Spotify controller is moved to spotcast so this can be closed

shocklateboy92:
Thanks!

PR #525, opened 2021-08-23
Adds controller for DashCast Unlimited fork

raulgbcr:
This adds the controller for the DashCast Unlimited fork (https://github.com/raulgbcr/dashcast-unlimited)

This fork forces the disableIdleTimeout command to allow the cast to not timeout on idle, since DashCast seems unmaintainned new ChromecastApp with new code and url have been created with the fork.

This was added for personal use, but figured out someone could find it usefull (https://github.com/skorokithakis/catt/issues/353). Please, take in mind that the name and style of the fork may be changed in the future if more features are added.

Regards


balloob:
Since this is for personal use, I suggest we don't merge this right now. It can be adopted in the future if your fork survives.

PR #542, opened 2021-10-05
Increase timeout for HostBrowser's polling

emontnemery:
Increase timeout for HostBrowser's polling

Related issue: https://github.com/home-assistant/core/issues/56793


emontnemery:
Set to draft as it seems Google has fixed this

PR #554, opened 2021-11-08
Instantiate metadataType when metadata is provided

Shane-Wiseman-Bose:
from testing and general usage, chromecast expects MetadataType to be provided when passing other metadata fields.

Given callers can either pass minor metadata fields such as title/thumb or a fully custom defined metadata dict, this change simple checks and instantiates the MetadataType field if it's not provided and user wants to provide media metadata. 

Proposal, allow the passing of the metadatatype via explicit function argument in case users' use case require the setting of this field.

### DIFF
--- a/pychromecast/controllers/media.py
+++ b/pychromecast/controllers/media.py
@@ -594,6 +594,12 @@ class MediaController(BaseController):
                 media["metadata"]["images"] = []
 
             media["metadata"]["images"].append({"url": thumb})
+
+        # Need to set metadataType if not specified
+        # https://developers.google.com/cast/docs/reference/messages#MediaInformation
+        if media["metadata"] and "metadataType" not in media["metadata"]:
+            media["metadata"]["metadataType"] = METADATA_TYPE_GENERIC
+
         if subtitles:
             sub_msg = [
                 {
@@ -612,11 +618,6 @@ class MediaController(BaseController):
                 "edgeType": "OUTLINE",
                 "edgeColor": "#000000FF",
             }
-
-        # if user has specified metadata but not specified metadataType
-        # https://developers.google.com/cast/docs/reference/messages#MediaInformation
-        if media["metadata"] and "metadataType" not in media["metadata"]:
-            media["metadata"]["metadataType"] = METADATA_TYPE_GENERIC
 
         if enqueue:
             msg = {
--- a/pychromecast/controllers/media.py
+++ b/pychromecast/controllers/media.py
@@ -612,12 +612,12 @@ class MediaController(BaseController):
                 "edgeType": "OUTLINE",
                 "edgeColor": "#000000FF",
             }
-           
+
         # if user has specified metadata but not specified metadataType
         # https://developers.google.com/cast/docs/reference/messages#MediaInformation
         if media["metadata"].keys() and "metadataType" not in media["metadata"]:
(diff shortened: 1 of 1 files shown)


emontnemery:
Can you give a clearer explanation of what problem this solves, ideally with an example

Shane-Wiseman-Bose:
according to cast documentation, when providing metadata, the metadata struct should include a metadataType field.

https://developers.google.com/cast/docs/reference/messages#MediaData

The current class provides the const variable but never initializes/utilizes it when constructing the media message.

https://github.com/home-assistant-libs/pychromecast/blob/7688ec6130dd0dc0e29023923b74f78d7cbfb563/pychromecast/controllers/media.py#L39

https://github.com/home-assistant-libs/pychromecast/blob/7688ec6130dd0dc0e29023923b74f78d7cbfb563/pychromecast/controllers/media.py#L579

emontnemery:
I understand what the patch does, but not what problem it solves.

You mention that:
> from testing and general usage, chromecast expects MetadataType to be provided when passing other metadata fields

What happens when `metadataType` is not set in the metadata object?

pychromecast/controllers/media.py
```diff
@@ -613,6 +613,11 @@ def _send_start_play_media(  # pylint: disable=too-many-locals
                 "edgeColor": "#000000FF",
             }
 
+        # if user has specified metadata but not specified metadataType
+        # https://developers.google.com/cast/docs/reference/messages#MediaInformation
+        if media["metadata"].keys() and "metadataType" not in media["metadata"]:
+            media["metadata"]["metadataType"] = METADATA_TYPE_GENERIC
+
```
emontnemery, line 620: Let's move this up to where the other metadata mangling is happening.

pychromecast/controllers/media.py
```diff
@@ -613,6 +613,11 @@ def _send_start_play_media(  # pylint: disable=too-many-locals
                 "edgeColor": "#000000FF",
             }
 
+        # if user has specified metadata but not specified metadataType
+        # https://developers.google.com/cast/docs/reference/messages#MediaInformation
+        if media["metadata"].keys() and "metadataType" not in media["metadata"]:
```
emontnemery, line 618: Why do we need to check if metadata's keys is truthy, isn't this enough:
```suggestion
        if "metadataType" not in media["metadata"]:
```

Shane-Wiseman-Bose:
> I understand what the patch does, but not what problem it solves.
> 
> You mention that:
> 
> > from testing and general usage, chromecast expects MetadataType to be provided when passing other metadata fields
> 
> What happens when `metadataType` is not set in the metadata object?

in my circumstance, when a device has screen, metadata is not applied (album art/title/etc). When looking at the logs, cast complains the metadataType field is null.

pychromecast/controllers/media.py
Shane-Wiseman-Bose: because metadata is initialized as an empty dict. We only want to set the MetadataType if the user has defined other metadata attributes, unless that is counter to the intention.

emontnemery: OK, got it.
An empty dict is falsy, there's no need to check the keys, I updated the suggestion accordingly.

emontnemery (approved):
Thanks, @Shane-Wiseman-Bose 👍

PR #559, opened 2021-11-16
HA Lovelace controller: Prevent sending connect message twice

bramkragten:
We would send a connect message twice, causing the app to lose its connection when creating a new one, and commands to fail.

When you request the view to show, it sends a connect message, then the app starts, it requests a status and sends another connect message, because we are not connected yet

### DIFF
--- a/pychromecast/controllers/homeassistant.py
+++ b/pychromecast/controllers/homeassistant.py
@@ -74,7 +74,8 @@ class HomeAssistantController(BaseController):
     def connect_hass(self, callback_function=None):
         """Connect to Home Assistant."""
         if self.hass_connected:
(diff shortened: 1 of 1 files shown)


pychromecast/controllers/homeassistant.py
```diff
@@ -69,7 +73,16 @@ def receive_message(self, _message, data: dict):
 
     def connect_hass(self, callback_function=None):
         """Connect to Home Assistant."""
+        if self.hass_connected:
+            callback_function()
```
balloob, line 77: We should test `callback_function` is defined.
```suggestion
            if callback_function:
                callback_function()
```

PR #571, opened 2021-12-08
Bump protobuf to 3.19.1 and regenerate protobuf files

emontnemery:


emontnemery:
I'm not sure if we want this, the problem it solves was caused by a Home Assistant custom component.

balloob:
I think that it can't hurt to bump protobuf to a newer version?

PR #581, opened 2021-12-29
Update README.rst

strunker:
So I had some pretty big blockers using the new library and discovery methods. I also finally realized what kind of firewall rules need to be created in order for zeroconf discovery to work properly. On windows I created two rules. one inbound, one outbound, for ports 1900,5353 and allowed them through the firewall. Afterwards I was able to do proper chromecast discovery without needing to disable my firewall. I also added some additional examples on how to discover numerous devices by friendly name, and I think a super important thing to include is the param discovery_timeout. I was getting many more devices using get_chromecasts() that I was not getting with get_listed_chromecasts() so I examined the difference and realized the timing is much more aggressive by default on get_listed_chromecasts(), the param helps alleviate that, and accounts for networks\devices that are slower to respond.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -154,12 +154,14 @@ to the IGNORE\_CEC list in PyChromecast like in the example below.
     pychromecast.IGNORE_CEC.append('*')  # Ignore CEC on all devices
     pychromecast.IGNORE_CEC.append('Living Room')  # Ignore CEC on Chromecasts named Living Room
 
-Firewall Rules
+Networking requirements
 -----------------
-MDNS discovery takes place on port UDP 5353.
-SSDP discovery takes place on port UDP 1900.
+Pychromecast relies on mDNS to discover cast devices. The mDNS protocol relies on multicast UDP on port 5353 which comes with several implications for discovery to work:
+- Multicast UDP must be forwarded by WiFI routers; some WiFi routers are known to drop multicast UDP traffic.
+- The device running pychromecast must allow both inbound and outbound traffic on port 5353.
+- The device running pychromecast must be on the same subnet as the cast devices because mDNS packets are not routed across subnets.
 
-You may want\need to create a firewall rule that allows both of these ports both inbound and outbound. Device discovery may not function properly until this rule is in place.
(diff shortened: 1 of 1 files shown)


strunker:
Anyone out there who can review this? This really should be merged in the readme document doesnt really provide everything you need to know to be successful in setting things up.

README.rst
```diff
@@ -144,6 +154,13 @@ to the IGNORE\_CEC list in PyChromecast like in the example below.
     pychromecast.IGNORE_CEC.append('*')  # Ignore CEC on all devices
     pychromecast.IGNORE_CEC.append('Living Room')  # Ignore CEC on Chromecasts named Living Room
 
+Firewall Rules
+-----------------
+MDNS discovery takes place on port UDP 5353.
+SSDP discovery takes place on port UDP 1900.
+
+You may want\need to create a firewall rule that allows both of these ports both inbound and outbound. Device discovery may not function properly until this rule is in place.
+
```
emontnemery, line 163:
```suggestion
Networking requirements
-----------------
Pychromecast relies on mDNS to discover cast devices. The mDNS protocol relies on multicast UDP on port 5353 which comes with several implications for discovery to work:
- Multicast UDP must be forwarded by WiFI routers; some WiFi routers are known to drop multicast UDP traffic.
- The device running pychromecast must allow both inbound and outbound traffic on port 5353.
- The device running pychromecast must be on the same subnet as the cast devices because mDNS packets are not routed across subnets.

If not all of these conditions are met, discovery will not work. In cases where these conditions are impossible to meet, it's possible to pass a list of known IP-addresses or host names to the discovery functions.

```

emontnemery (commented):
pychromecast does not rely on SSDP for discovery, so we don't need to mention that. If you don't agree @strunker, please open another PR which motivates why SSDP is required.

emontnemery (approved):
Thanks @strunker 👍 

I'm sorry about the very long delay for reviewing this.

strunker:
Thanks @emontnemery  all good on the delay. Hope these additions help others who run on Windows.

Issue #580, opened 2021-12-29
"Failed to determine cast type"


strunker:
I am receiving the below error during get_chromecasts() method call. I have never seen this error prior to moving to the latest 10.2.2.

The discovery ultimately works, but it takes quite a while to complete. I understand there are other methods for discovery in these later versions but this get_chromecasts has proven to be more reliable in my experience so I would prefer to continue to use it if possible.

Mainly just wanted to report this odd behavior as I cant make sense of what "failed to determine cast type" means, and it repeats twice before the discovery completes. 

![image](https://user-images.githubusercontent.com/6154660/147632847-ad23057d-ff29-4ceb-af7d-7932074ba51f.png)

elibroftw:
I'm certain https://github.com/home-assistant-libs/pychromecast/blob/master/pychromecast/dial.py#L141 is responsible for the log output. However, the reasoning is not clear so I suggest that the library should log the reason behind not being able to determine the cast type rather than shrug its shoulders.

For now I'm going to fork the library, and do some debugging to get to the bottom of this issue as my users are encountering this problem.

emontnemery:
This error means the chromecast device does not allow fetching basic status via HTTP. The logic around this has not intentionally been changed in pychromecast, is it really an issue in pychromecast 10.2.2, and not an issue with the cast device?

What happens if you do this from command line:
```
curl -k "https://<chromecast_ip>:8443/setup/eureka_info?params=device_info,name" | python3 -m json.tool
curl -k "http://<chromecast_ip>:8008/setup/eureka_info?params=device_info,name" | python3 -m json.tool
```

emontnemery:
@elibroftw you mentioned in a comment you have forked pychromecast, have you worked around this problem then?

elibroftw:
> @elibroftw you mentioned in a comment you have forked pychromecast, have you worked around this problem then?

It was a false psotiive. The "problem" is probably due to the device going offline. I thought it was a problem because I didn't expect devices to be going offline but then realized it was a race condition I didn't account for.

strunker:
> > @elibroftw you mentioned in a comment you have forked pychromecast, have you worked around this problem then?
> 
> It was a false positive. The "problem" is probably due to the device going offline. I thought it was a problem because I didn't expect devices to be going offline but then realized this was a race condition I didn't account for.
> 
> This is at least for my case! Not sure about everyone else's.

[...]

PR #582, opened 2021-12-29
Update dial.py

strunker:
cast type errors from zeroconf, for starters, arent very verbose so printing warnings for them doesnt really add anything except more messages in the client. I would propose relegating these over to the debug channel instead of the warning channel so they dont show up in the console anymore. The errors are handled, and the cast type is overridden anyway and set to CAST_TYPE_CHROMECAST, so it seems kind of silly to warn on these errors. 

At a minimum it would be nice if the error text was included in the warning log so that the user has some type of idea what is going on and how they can potentially fix the problem. I am not even really personally clear on what the problem is I didnt bother to log the actual error, instead I just cut it over to debug channel and let it be.

### DIFF
--- a/pychromecast/dial.py
+++ b/pychromecast/dial.py
@@ -138,7 +138,7 @@ def get_cast_type(cast_info, zconf=None, timeout=30, context=None):
             _LOGGER.debug("cast type: %s, manufacturer: %s", cast_type, manufacturer)
 
         except (urllib.error.HTTPError, urllib.error.URLError, OSError, ValueError):
-            _LOGGER.warning("Failed to determine cast type")
+            _LOGGER.debug("Failed to determine cast type")
             cast_type = CAST_TYPE_CHROMECAST
 
     return CastInfo(


emontnemery:
@strunker the error is unexpected which is why it's logged as a warning.
Do you actually see this in the log, if you do would you mind logging the exception and paste the output here?

strunker:
> @strunker the error is unexpected which is why it's logged as a warning. Do you actually see this in the log, if you do would you mind logging the exception and paste the output here?

Can you provide instructions for exactly how you want me to log it please. And no where I am seeing the logging messages is directly in my console view. I run a python script that has chromecast integration on a windows machine, all of which runs from a Flask server, so I see all the output in a CMD window as all of this stuff runs continuously. I never saw these errors prior, and they dont offer any explanation, until this new version of pychromecast. Additionally, all of the devices ultimately end up discovering anyway, so im not really sure what its referring to.

emontnemery:
#586 does the opposite of this PR by printing the exception instead of decreasing he severity. Could you reproduce the problem on latest master and share a log in #580 ?

PR #585, opened 2022-01-07
Add customData to play_media method

tibidi:
This code allows passing custom data when needed.


tibidi:
Need more investigations

PR #586, opened 2022-01-09
Log error code on exception

ColinRobbins:
When there is an error getting the cast type, log the error condition, to enable furher debugging.

### DIFF
--- a/pychromecast/dial.py
+++ b/pychromecast/dial.py
@@ -137,8 +137,8 @@ def get_cast_type(cast_info, zconf=None, timeout=30, context=None):
                 cast_type = CAST_TYPE_AUDIO
             _LOGGER.debug("cast type: %s, manufacturer: %s", cast_type, manufacturer)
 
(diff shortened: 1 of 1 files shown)


emontnemery:
Thanks, @ColinRobbins 👍

emontnemery (approved):

emontnemery:
I guess this is related to #580? Did you get anywhere by printing the exception?

elibroftw:
`repr(err)` would be more informative. Case in point:
```py
>>> e = Exception('hello world')
>>> str(e)
'hello world'
>>> repr(e)
"Exception('hello world')"
```

emontnemery:
If we really want to print all the exception details we should instead do:
```py
_LOGGER.warning("Failed to determine cast type", exc_info=True)
```

elibroftw:
I think exec info is an exception not a boolean.

emontnemery:
It's either or: https://docs.python.org/3/library/logging.html#logging.Logger.debug

PR #598, opened 2022-03-01
add plex support for previous and next

guirem:
Currently, cast device playing plex media returns basic/default supported_media_commands data that does not reflect what the cast device can do/accept.
In hass, the next and previous command is not available.

This PR will activate the next/previous command for plex media_player by overwriting the supported media information received by the cast device.

### DIFF
--- a/pychromecast/controllers/plex.py
+++ b/pychromecast/controllers/plex.py
@@ -135,6 +135,7 @@ def media_to_chromecast_command(
                 "directStream": directStream,
                 "subtitleSize": subtitleSize,
                 "audioBoost": audioBoost,
+                "providerIdentifier": "com.plexapp.plugins.library",
                 "server": {
                     "machineIdentifier": machineIdentifier,
                     "transcoderVideo": transcoderVideo,
@@ -283,6 +284,14 @@ class PlexController(BaseController):
         """Send next command."""
         self._send_cmd({MESSAGE_TYPE: TYPE_NEXT})
 
+    def queue_next(self):
+        """Send QUEUE_NEXT as NEXT command."""
+        self.next()
+
+    def queue_prev(self):
+        """Send QUEUE_PREV as NEXT command."""
+        self.previous()
+
     def seek(self, position, resume_state="PLAYBACK_START"):
         """Send seek command.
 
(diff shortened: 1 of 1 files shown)


emontnemery:
Why is the plex app reporting the wrong information? Is there any way to know when the app will accept next/previous and when it won't?

guirem:
I believe it can relate to these already known issues https://github.com/home-assistant/core/issues/58888  and https://github.com/home-assistant/core/issues/49243 .
Problem seems to be in google cast plex app implementation. 

I tested with several type of media and configuration using official desktop player with chromecast broadcast and the device retruns the same supported media command code.

I guess next/previous won't be possible if this is a single item playing but not even sure of that as it always seem to be creating some sort of virtual playlist (plex player or via hass).

guirem:
Hi, 
Any chance this change may be in a future release?
@emontnemery what additional info I could provide ?

pychromecast/controllers/plex.py
```diff
@@ -355,9 +364,13 @@ def status(self):
         Returns:
             pychromecast.controllers.media.MediaStatus: Slightly modified status with patched
                                                         method for episode_title.
+                                                        Overwrite supported_media_commands 
+                                                        when set to MEDIA_BASIC
         """
         status = self._socket_client.media_controller.status
         status.episode_title = episode_title
+        if status.supported_media_commands == 12303: 
+            status.supported_media_commands = 12543
```
emontnemery, line 373: Let's not use magic numbers, use the `CMD_SUPPORT_*`-constants defined in media.py instead.

The second row is doing:
```py
    status.supported_media_commands |= CMD_SUPPORT_SKIP_FORWARD | CMD_SUPPORT_SKIP_BACKWARD | CMD_SUPPORT_QUEUE_NEXT | CMD_SUPPORT_QUEUE_PREV
```

The first row seems very specific, is it correct, or should we always set the bits indicating support for previous and next?

pychromecast/controllers/plex.py
```diff
@@ -283,6 +284,14 @@ def next(self):
         """Send next command."""
         self._send_cmd({MESSAGE_TYPE: TYPE_NEXT})
 
+    def queue_next(self):
+        """Send QUEUE_NEXT as NEXT command."""
```
emontnemery, line 288: Let's explain why we need to do this.

pychromecast/controllers/plex.py
```diff
@@ -283,6 +284,14 @@ def next(self):
...
+        self.next()
+
+    def queue_prev(self):
+        """Send QUEUE_PREV as NEXT command."""
+        self.previous()
+
```
emontnemery, line 294: Does Home Assistant really end up calling these functions; have you tested it?
The way I understand it, Home Assistant will still call `MediaController.queue_next`.

emontnemery:
I'm closing this PR now since the review comments have not been addressed in almost two years. @guirem Please open a new PR addressing my review comments if you still want to see this functionality merged.

PR #607, opened 2022-03-30
Refactor media controllers

emontnemery:
Refactor media controllers to make it easier to implement support for customized media player apps which load media via the standard namespace `urn:x-cast:com.google.cast.media`.

### DIFF
--- a/pychromecast/controllers/__init__.py
+++ b/pychromecast/controllers/__init__.py
@@ -1,16 +1,28 @@
 """
 Provides controllers to handle specific namespaces in Chromecast communication.
 """
+from __future__ import annotations
+
 import abc
 import logging
+from typing import TYPE_CHECKING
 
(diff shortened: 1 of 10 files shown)


balloob (approved):

emontnemery:
@blawford Could you help verify the BBC iplayer still works with this change; I can't test it myself because everything seems to be region locked to the UK.

It would be great if you could test:
- Play a clip with no app running on the chromecast
- Play a clip with the BBC iPlayer app already running on the chromecast
- Play a clip with another media receiver based app already running on the chromecast, for example BBC sounds
- Play a clip with a non media receive based app already running on the chromecast, for example youtube

You can use the sample scripts for testing; examples/bbciplayer_example.py, examples/bbcsounds_example.py, examples/youtube_example.py

PR #613, opened 2022-04-11
Include URL and UUID in Home Assistant controller messages

emontnemery:
Can't be merged until https://github.com/home-assistant/frontend/pull/12354 is merged and an updated app is published

### DIFF
--- a/pychromecast/controllers/homeassistant.py
+++ b/pychromecast/controllers/homeassistant.py
@@ -17,6 +17,7 @@ class HomeAssistantController(BaseController):
     def __init__(
         self,
         hass_url,
+        hass_uuid,
         client_id,
         refresh_token,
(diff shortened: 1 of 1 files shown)


pychromecast/controllers/homeassistant.py
```diff
@@ -45,6 +47,7 @@ def hass_connected(self):
             self.status is not None
             and self.status["connected"]
             and self.status["hassUrl"] == self.hass_url
+            and self.status["hassUUID"] == self.hass_uuid
```
balloob, line 50: Why is this necessary? In what case are the URLs the same but the UUIDs not?

emontnemery: When developing or testing with multiple hass instances this can be the case.

balloob (approved):

PR #620, opened 2022-05-05
Lookup known cast models in mDNS discovery

emontnemery:


balloob (approved):

pychromecast/discovery.py
```diff
@@ -206,8 +208,11 @@ def get_value(key):
 
         # Lock because the HostBrowser may also add or remove items
         with self._services_lock:
-            cast_type = CAST_TYPE_GROUP if service.port != 8009 else None
-            manufacturer = "Google Inc." if service.port != 8009 else None
+            if service.port != 8009:
+                cast_type = CAST_TYPE_GROUP
+                manufacturer = MF_GOOGLE
+            else:
+                cast_type, manufacturer = CAST_TYPES.get(model_name.lower(), None, None)
```
balloob, line 215: `get` only takes 2 arguments and this are 3.

PR #625, opened 2022-05-12
pin to protobuf v3

chapmanjacobd:
It seems like protobuf v4 is causing issues


emontnemery (approved):
Thanks, @chapmanjacobd 👍

PR #629, opened 2022-05-26
Initial submission for multi-screen YouTube cast

loganimal2021:
Submitting example .py that allows casting of YouTube videos to multiple selected screens. Still see some issues with synchronized timing of casting, but generally pretty reliable.


emontnemery:
I'm not sure how useful an example this is considering it "just" starts playback on the same youtube clip on multiple devices with no synchronization.

emontnemery:
I'm closing this PR now since the question above is not answered. @loganimal2021 Please open a new PR if you still think your example is useful.

PR #646, opened 2022-07-04
Add support for DRTV app as a custom controller

mchro:
Notably, the app requires two JWTs ("session tokens") to allow playback. These
can either be retrieved from a browser and given (expiry seem to not be
critical), or alternatively retrieval of an anonymous token using Selenium will
be attempted.


pychromecast/controllers/drtv.py
```diff
@@ -0,0 +1,118 @@
...
+                "senderDeviceType": "pyChromeCast",
+                "showDebugOverlay": False,
+                "userId": ""
+            },
+        }
+
+        print(msg)
+        self.send_message(msg, inc_session_id=True, callback_function=callback_function)
+
+    def _get_drtokens(self):
```
mchro, line 70: Guidance on whether this approach to retrieving tokens is acceptable is welcome :-)
It should work with Selenium as an optional dependency in this way.

mchro: The stated lifetime of the tokens is 12 hours; there seems to be _some_ grace period, but I have been unable to ascertain how much.
I'm mostly worried about the user experience, if having to provide a hardcoded token that then regularly expires. Of course we could leave it up to the consumer of this API, i.e. somewhere inside Home Assistant, to do the token maintenance?

emontnemery: Maybe code for getting the token etc. would fit in well in https://github.com/Eerovil/castextensions, @Eerovil wdyt?

mchro: In the meantime I have pushed a commit to make the tokens a mandatory parameter in the controller, but left the Selenium way in the example code.
I also dug a bit deeper in the JWTs, and found a parameter in there indicating the tokens are valid for 1 month.

emontnemery:
> but left the Selenium way in the example code

That's fine.
Unless it's trivial, please amend the example with instructions on how to setup selenium and so on.

mchro:
> > but left the Selenium way in the example code
> 
> That's fine. Unless it's trivial, please amend the example with instructions on how to setup selenium and so on.

Should be pretty trivial, just needs the Python module `selenium` and Chromium. I'll link the official module.

Issue #290, opened 2019-05-04
version parameter in setup.py out of sync with your __version__ in pychromecast/__init__.py


sebma:
Hi,

The [`version` parameter in your `setup.py`](../blob/master/setup.py#L8) is out of sync with your [`__version__` in your `pychromecast/__init__.py`](../blob/master/pychromecast/__init__.py#L20) : 

```bash
$ grep -P "version.*=.*\d" setup.py pychromecast/__init__.py 
setup.py:    version='3.2.1',
pychromecast/__init__.py:__version_info__ = ('0', '7', '6')
```

Here a proposition to have them always in sync (there are other techniques at https://packaging.python.org/guides/single-sourcing-package-version/) : 

You can create a `pychromecast/version.py` with : 
```python
__version_info__ = ('3', '2', '1')
__version__ = '.'.join(__version_info__)
```

and modify your `setup.py` like this : 

```python
exec(open('pychromecast/version.py').read())

setup(
    name='PyChromecast',
    version=__version__,
```

and finally add that in your `pychromecast/__init__.py` : 

`from pychromecast.version import __version__`

PR #647, opened 2022-07-07
Drop __version__

emontnemery:
~~This implements the suggestion from #290~~

This PR drops the `__version__` from pychromecast


balloob:
We can start with this. I think a better follow-up would be to just remove it.

balloob:
Actually let's just remove it in this PR.

frenck (approved):

PR #676, opened 2022-12-02
Fix error with select when python process exceeds 1024 open file descriptors

logan893:
When the python process has over 1024 open file descriptors, the select.select() function will throw an error.
Error in select call: filedescriptor out of range in select()

Replace select.select with select.poll to avoid this error.

https://github.com/home-assistant/core/issues/79519

### DIFF
--- a/pychromecast/socket_client.py
+++ b/pychromecast/socket_client.py
@@ -565,7 +565,8 @@ class SocketClient(threading.Thread):
 
         # poll the socket, as well as the socketpair to allow us to be interrupted
         rlist = [self.socket, self.socketpair[0]]
-        # Map file descriptors to socket objects
+        # Map file descriptors to socket objects because select.select does not support fd > 1024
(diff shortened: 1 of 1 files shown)


pychromecast/socket_client.py
```diff
@@ -565,8 +565,14 @@ def run_once(self, timeout=POLL_TIME_NON_BLOCKING):
 
         # poll the socket, as well as the socketpair to allow us to be interrupted
         rlist = [self.socket, self.socketpair[0]]
+        # Map file descriptors to socket objects
```
emontnemery, line 568:
```suggestion
        # Map file descriptors to socket objects because select.select does not support fd > 1024
        # https://stackoverflow.com/questions/14250751/how-to-increase-filedescriptors-range-in-python-select
```

emontnemery (approved):
Thanks, @logan893 👍

PR #680, opened 2022-12-07
Adding support for reported unknown devices

SplinterHead:
There are a list of unknown devices in the homassistant/core issues, this PR gives the library the support for those.

It also creates constants for all the manufacturer names


emontnemery:
Thanks a lot @SplinterHead 👍

PR #690, opened 2022-12-21
Fix poll/select for Windows

strunker:
Added OS detection in socket code. If os is not windows, the newer select poll code is used, otherwise the older working Windows code is utlized.


elibroftw:
@emontnemery Can we get this merged / reviewed? I'm currently shipping my music player with 13.0.1 after realizing the 13.0.2 I shipped two days ago was buggy for my users.

ganlbarone:
I dont see why you cant. The fix is pretty straight forward and works well. I have been using this fork for months on end without a single issue.

Issue #665, opened 2022-10-12
PyChromecast how to use page incorrectly shows "device" in examples


benbaez:
Since device was removed in 10.0.0, these should be changed from:

>> [cc.device.friendly_name for cc in chromecasts]

>> print(cast.device)

to

>>[cc.cast_info.friendly_name for cc in chromecasts]

>> print(cast.cast_info)

Fluffik3666:
haha ran into the same issue - lifesaver:) they should really change this, it's miss leading and doesn't help new users starting to use this library!

PR #737, opened 2023-07-21
updating the examples in readme so the they don't throw errors about device not existing

sophiathekitty:
fixed the cc.device.friendly name to use cc.cast_info.friendly_name and removed the print(cast.device) because Chromecast objects don't have device anymore. saw that this has been an issue since at least last October #665


emontnemery:
Thanks @sophiathekitty 👍 

This was however also fixed by #751

FixThisB:
Example is still broken, please merge the solutions provided here. 

If you want to keep the sanity check output maybe include `print(cast.cast_info)` instead of scrapping `print(cast.device)` all together

PR #751, opened 2023-10-10
Update README.rst to reflect refactoring of cast.device to cast.cast_info

davidjabon:
Post version 10, "device" has been refactored to something similar "cast_info."  Readme for current version should reflect this change, because the code will not run.


emontnemery (approved):
Thanks, @davidjabon 👍

elibroftw:
Shouldn't this be a major change according to semver and not a minor? Just asking because I usually use ~=13.0 for pychromecast and the only reason I caught this is because I get all release notifications and saw this as a change.

PR #758, opened 2023-10-27
Create ShakaPlayer Controller

juanfadev:
ShakaPlayer is a much more versatile and improved player over the default one. It allows for HLS and DASH play with custom arguments  such as DRM. I have included an example of this DRM usage using ClearKey but Widevine or others can be used.


juanfadev:
@balloob could you please check this PR or get someone to check it? I think it will improve a lot the posibilities for playing any media.

examples/shaka_drm_example.py
```diff
@@ -0,0 +1,131 @@
+"""
+Example on how to use the BubbleUPNP Controller to play an URL.
```
emontnemery, line 2: Stale docstring

emontnemery:
Is the shaka player documented somewhere, what arguments does it accept and so on?

Lamarqe:
Google says, they will switch to Shaka player as the default media player as soon as its mature enough:
https://developers.google.com/cast/docs/web_receiver/shaka_migration

> The Web Receiver SDK will transition to using Shaka Player as the default player for HLS content playback. This will be done in a phased approach to guide partners through the migration process.
> 1. Oct '22 	The Cast Web Receiver SDK introduces APIs to opt-in to play HLS content on Shaka Player.
> 2. H2 '24 	The Cast Web Receiver SDK changes the default player to Shaka Player for HLS content playback.
> 

I dont think pychromecast needs to be extended while shaka player is still in this beta-ish phase.
If you want to use it, you can still do it, based on a custom controller.

examples/shaka_drm_example.py
emontnemery, line 2:
```suggestion
Example on how to use the Shaka Controller to play an URL.

```

PR #770, opened 2023-12-01
Increase homeassistant lovelace connect timeout and make it configurable

cicchr:
I was having an issue where homeassistant dashboards would not always load correctly when casting. I also discovered I wasn't the only one experiencing this: https://github.com/home-assistant/core/issues/82823. I tracked down the source of the warning message and did some testing and confirmed the exceptions were always thrown after 10 seconds exactly. Increasing the timeout allows the dashboard to show correctly and the error messages are gone.

10 seconds seems somewhat arbitrary here and some cast devices can be slow to load. Also since a bunch of people are having this issue I think it makes sense to increase the default timeout.

Another option would be to make this configurable in the cast integration to allow users experiencing this issue to set a larger timeout.

### DIFF
--- a/pychromecast/controllers/homeassistant.py
+++ b/pychromecast/controllers/homeassistant.py
@@ -9,6 +9,7 @@ from . import BaseController
 
 
 APP_NAMESPACE = "urn:x-cast:com.nabucasa.hast"
+DEFAULT_HASS_CONNECT_TIMEOUT = 30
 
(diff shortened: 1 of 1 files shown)


smarthomefamilyverrips (approved):

emontnemery (approved):
Thanks, @cicchr 👍

PR #779, opened 2024-01-15
Make optional callback functions kwarg only

emontnemery:
## Breaking change

Most user facing functions accepting an optional callback function have been changed such that the optional arguments are not kwarg only

## Change

Make optional callback functions kwarg only

## Background

This is the first step in refactoring the callback functions to accept a bool status and an optional response.

Refactoring the callback handling is needed because callbacks are currently only called if the request suceeds; there's no call if the request fails. This means any user relying on callbacks will hang forever or until a timeout when the request fails.

### DIFF
--- a/pychromecast/__init__.py
+++ b/pychromecast/__init__.py
@@ -381,7 +381,9 @@ class Chromecast:
         """Start an app on the Chromecast."""
         self.logger.info("Starting app %s", app_id)
 
-        self.socket_client.receiver_controller.launch_app(app_id, force_launch)
+        self.socket_client.receiver_controller.launch_app(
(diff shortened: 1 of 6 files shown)


MartinHjelmare (approved):
Is this a breaking change for users of the library?

What's the reason for the change?

> callback functions to accept a bool status and an optional response

emontnemery:
> Is this a breaking change for users of the library?
> 
> What's the reason for that goal?
> 
> > callback functions to accept a bool status and an optional response

Thanks Martin, the PR has been updated with this information

MartinHjelmare:
I fixed a typo.

PR #782, opened 2024-01-15
Remove impossible exception handling in youtube controller

emontnemery:
The comment for the removed exception handling claims:
> This gets the screenId but always throws

However, this is impossible, there's no case where we send a message and after that raise `UnsupportedNamespace`

### DIFF
--- a/pychromecast/controllers/youtube.py
+++ b/pychromecast/controllers/youtube.py
@@ -83,11 +83,7 @@ class YouTubeController(BaseController):
         (send message will launch app if it is not running).
         """
         self.status_update_event.clear()
-        # This gets the screenId but always throws. Couldn't find a better way.
-        try:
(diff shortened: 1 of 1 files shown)


MartinHjelmare:
The base controller can raise. Will that path never be taken?

emontnemery:
> The base controller can raise. Will that path never be taken?

Sure, that path can be taken, but it's not expected and we should not catch and ignore it.

The code I want to remove have this comment:
```py
# This gets the screenId but always throws. Couldn't find a better way.
```

This is impossible, as explained in the PR description and on Discord; there's no case where `BaseController.send_message` first sends a message and after that raises UnsupportedNamespace. Maybe there was some other bug in pychromecast when that code was added forcing the author to add the workaround.

MartinHjelmare (approved):
Ok.

PR #783, opened 2024-01-16
Change callbacks to be passed status and result

emontnemery:
## Breaking change

The signature of callbacks called when requests are completed has been changed:
- The callbacks must now accept a bool status and an optional dict with response
- The callbacks will be called both in case the request succeeds and when it doesn't

New callback signature:
```py
CallbackType = Callable[[bool, dict | None], None]
"""Signature of optional callback functions supported by methods sending messages.

The callback function will be called with a bool indicating if the message was sent
and an optional response dict.
"""
```

## Change

Callback functions are now called not only on success but also on error. This prevents failing requests from hanging forever. Also, many user facing functions which send messages to the remote device but do not accept a callback now accept a timeout parameter.

Background in PR https://github.com/home-assistant-libs/pychromecast/pull/779

Needs https://github.com/home-assistant-libs/pychromecast/pull/784

### DIFF
--- a/pychromecast/__init__.py
+++ b/pychromecast/__init__.py
@@ -23,9 +23,10 @@ from .discovery import (  # noqa: F401
     stop_discovery,
 )
 from .dial import get_cast_type
-from .const import CAST_TYPE_CHROMECAST, SERVICE_TYPE_HOST
+from .const import CAST_TYPE_CHROMECAST, REQUEST_TIMEOUT, SERVICE_TYPE_HOST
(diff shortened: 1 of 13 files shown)


MartinHjelmare:
The breaking change sentence sounds weird. It sounds like the callback is called when a request has changed signature.

MartinHjelmare:
I think you should document what the callback parameters mean.

MartinHjelmare (approved):

PR #784, opened 2024-01-16
Require Python 3.11+

emontnemery:
Run tests with Python 3.11 and update Readme to mention Python 3.11+ is required

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,6 +17,7 @@ setup(
     include_package_data=True,
     platforms="any",
     install_requires=list(val.strip() for val in open("requirements.txt")),
+    python_requires=">=3.11",
     classifiers=[
         "Intended Audience :: Developers",
         "License :: OSI Approved :: MIT License",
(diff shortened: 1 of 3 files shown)


MartinHjelmare (changes requested):
Set `python_requires=">=3.11"` in `setup.py` too.

setup.py
```diff
@@ -17,6 +17,7 @@
     include_package_data=True,
     platforms="any",
     install_requires=list(val.strip() for val in open("requirements.txt")),
```
MartinHjelmare, line 19: Side note: It's not good practice to set the production requirements the same way as the development requirements. I recommend pinning the development requirements strictly in requirements files (or an optional list in pyproject.toml) and have as forgiving pinning for the production as possible in requirements files or directly in setup.py (or pyproject.toml).

It would also be good to change to pyproject.toml at some point.

MartinHjelmare (approved):

PR #786, opened 2024-01-16
Make optional callback functions kwarg only in plex

emontnemery:
## Breaking change

User facing functions in the plex controller which accept an optional callback function have been changed such that the optional arguments are now kwarg only

## Change

Make optional callback functions kwarg only in dashcast

Background in PR https://github.com/home-assistant-libs/pychromecast/pull/779

### DIFF
--- a/pychromecast/controllers/plex.py
+++ b/pychromecast/controllers/plex.py
@@ -260,10 +260,10 @@ class PlexController(BaseController):
 
         return False
 
-    def update_status(self, callback_function_param=False):
+    def update_status(self, *, callback_function= None):
         """Send message to update status."""
         self.send_message(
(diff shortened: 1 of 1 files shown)


pychromecast/controllers/plex.py
```diff
@@ -260,10 +260,10 @@ def receive_message(self, _message, data: dict):
 
         return False
 
-    def update_status(self, callback_function_param=False):
+    def update_status(self, *, callback_function= None):
```
MartinHjelmare, line 263:
```suggestion
    def update_status(self, *, callback_function=None):
```

MartinHjelmare (approved):

PR #792, opened 2024-01-16
Add type annotations to dial.py

emontnemery:


pychromecast/dial.py
```diff
@@ -49,24 +57,30 @@ def get_host_from_service(service, zconf):
...
+) -> tuple[str | None, int | None]:
     """Get hostname or IP + port from zeroconf service_info."""
     host = None
     port = None
-    if (
-        service_info
-        and service_info.port
-        and (service_info.server or len(service_info.addresses) > 0)
-    ):
+    if service_info and service_info.port:
```
MartinHjelmare, line 66: There's no risk that the port is 0, right?

MartinHjelmare (approved):

pychromecast/dial.py
emontnemery: No, port 0 is not a valid port.

PR #793, opened 2024-01-17
Add type annotations to discovery.py

emontnemery:


pychromecast/discovery.py
```diff
@@ -116,26 +133,32 @@ def remove_service(self, _zconf, typ, name):
...
+            _LOGGER.debug("remove_service unknown %s, %s", type_, name)
             return
 
         if device_removed:
             self._cast_listener.remove_cast(uuid, name, cast_info)
         else:
             self._cast_listener.update_cast(uuid, name)
 
-    def update_service(self, zconf, typ, name):
+    def update_service(self, zc: zeroconf.Zeroconf, type_: str, name: str) -> None:
```
MartinHjelmare, line 144: Changing parameter names is a breaking change if the function is called with keyword arguments.

pychromecast/discovery.py
```diff
@@ -116,26 +133,32 @@ def remove_service(self, _zconf, typ, name):
...
         """Called by zeroconf when an mDNS service is updated."""
-        _LOGGER.debug("update_service %s, %s", typ, name)
-        self._add_update_service(zconf, typ, name, self._cast_listener.update_cast)
+        _LOGGER.debug("update_service %s, %s", type_, name)
+        self._add_update_service(zc, type_, name, self._cast_listener.update_cast)
 
-    def add_service(self, zconf, typ, name):
+    def add_service(self, zc: zeroconf.Zeroconf, type_: str, name: str) -> None:
```
MartinHjelmare, line 149: Below we call the parameters, `zconf` and `typ`. Why do we change here?

pychromecast/discovery.py
emontnemery: This class implements the `zeroconf.ServiceListener` interface: https://github.com/python-zeroconf/python-zeroconf/blob/dfc9b8d7dec519ca713a811613122718cb2d733e/src/zeroconf/_services/__init__.py#L37-L45, the parameter names were previously not aligned with the interface.

pychromecast/discovery.py
emontnemery: `_add_update_service` is not part of the `zeroconf.ServiceListener` interface, we can use any parameter name there.

pychromecast/discovery.py
MartinHjelmare: Ok.

MartinHjelmare (approved):

PR #796, opened 2024-01-17
Add type annotations to controllers/receiver.py

emontnemery:


pychromecast/controllers/receiver.py
```diff
@@ -15,8 +15,11 @@
     REQUEST_TIMEOUT,
     SESSION_ID,
 )
+from ..generated.cast_channel_pb2 import (  # pylint: disable=no-name-in-module
```
MartinHjelmare, line 18: Will this always be the case for generated modules that pylint doesn't understand them?

MartinHjelmare (approved):

pychromecast/controllers/receiver.py
emontnemery: It seems like it, yes.
Pylint actually very recently gained support for `pyi` stub files https://github.com/pylint-dev/pylint/issues/4987, but I guess the support is still rough around the edged.

For example, this comment mentions something very similar: https://github.com/pylint-dev/pylint/issues/4987#issuecomment-1821905002

PR #802, opened 2024-01-25
Remove Python 3.1 compatibility from socket client

emontnemery:
This was fixed in Python 3.2 by https://bugs.python.org/issue10272


MartinHjelmare (approved):

MartinHjelmare (commented):
Side note: `socket.timeout` is deprecated in favor of `TimeoutError`.

https://docs.python.org/3/library/socket.html#socket.timeout

PR #805, opened 2024-01-28
Add type annotations to socket_client.py

emontnemery:


pychromecast/socket_client.py
```diff
@@ -860,24 +882,21 @@ def send_message(
...
         msg.protocol_version = msg.CASTV2_1_0
         msg.source_id = self.source_id
         msg.destination_id = destination_id
-        msg.payload_type = (
-            cast_channel_pb2.CastMessage.STRING  # pylint: disable=no-member
-        )
+        msg.payload_type = CastMessage.STRING
         msg.namespace = namespace
-        msg.payload_utf8 = _json_to_payload(data)
+        msg.payload_utf8 = _json_to_payload(data)  # type: ignore[assignment]
```
emontnemery, line 899: The existing code works, but violates the typing of `CastMessage` which claims `CastMessage.payload_utf8` is `str`.

MartinHjelmare:
There's a merge conflict.

pychromecast/socket_client.py
```diff
@@ -273,7 +286,8 @@ def mdns_backoff(service, retry):
                     continue
                 try:
                     if self.socket is not None:
-                        self.socket.close()
+                        # If we retry connecting, we need to clean up the socket again
+                        self.socket.close()  # type: ignore[unreachable]
```
MartinHjelmare, line 290: Why is this unreachable?

emontnemery: I don't think it is, I think mypy gets confused because we set it to None here: https://github.com/home-assistant-libs/pychromecast/blob/518187629f4980504f1e3aaf1886040fa2f8ef5a/pychromecast/socket_client.py#L241

MartinHjelmare: Ok.

MartinHjelmare (approved):

PR #811, opened 2024-02-01
Add type annotations to controllers/homeassistant.py

emontnemery:


pychromecast/controllers/homeassistant.py
```diff
@@ -53,13 +58,13 @@ def __init__(
         #   lovelacePath?: string | number | null;
         #   urlPath?: string | null;
```
MartinHjelmare, line 59: What are these comments?

MartinHjelmare (approved):

pychromecast/controllers/homeassistant.py
emontnemery: It's the typescript definition of the message

MartinHjelmare: I'd add a comment explaining that.

PR #813, opened 2024-02-01
Add type annotations to controllers/supla.py

emontnemery:


pychromecast/controllers/supla.py
```diff
@@ -48,8 +55,12 @@ def play_media(self, media_id, is_live=False, callback_function=None):
             no_add_request_id=True,
         )
 
-    def quick_play(self, media_id=None, is_live=False, **kwargs):
+    def quick_play(
+        self, media_id: str | None = None, is_live: bool = False, **kwargs: Any
```
MartinHjelmare, line 59: Why is the media id optional if it's required?

emontnemery: It shouldn't be optional. I plan update all controllers which implement quick_play to make the `media_id` required in a follow-up PR.

MartinHjelmare (approved):

PR #820, opened 2024-02-06
Add overloads for get_chromecasts

emontnemery:


pychromecast/__init__.py
```diff
@@ -171,6 +171,33 @@ def get_chromecast_from_uuid(uuid: UUID) -> Chromecast:
...
+) -> tuple[list[Chromecast], CastBrowser]:
+    ...
+
+
+@overload
+def get_chromecasts(
+    tries: int | None = None,
+    retry_wait: float | None = None,
+    timeout: float | None = None,
+    *,
```
emontnemery, line 192: This is added to allow specifying the value of blocking, which is otherwise not possible.
More details here: https://github.com/python/mypy/issues/7333

emontnemery:
An alternative to the changes in this PR would be to do a breaking change where `get_chromecasts` always returns the same type. Considering the large amount of breaking changes in the upcoming release, it may be a better option.

frenck (approved):

frenck:
I can't judge the latter. But if there are lots of breaking things already, I can imagine that this one doesn't matter.

PR #822, opened 2024-02-06
Update requirements for development and testing

emontnemery:


requirements.txt
emontnemery, line 1: Maybe we should just remove this file and move this to requirements.txt?

requirements.txt
```diff
@@ -1,3 +1,3 @@
-protobuf>=4.25.1
-zeroconf>=0.25.1
-casttube>=0.2.0
+casttube==0.2.1
+protobuf==4.25.2
+zeroconf==0.131.0
```
emontnemery, line 3: I'm not sure if we want to pin to latest or to minimum required version when testing?

MartinHjelmare: For development, the idea is to pin strictly to the latest working versions. For production, the idea is to require a minimum version as the oldest working version.

Depending on the situation a maximum version can also be added for production requirements (normally the next major version). Production requirements should be as forgiving as possible.

requirements.txt
MartinHjelmare: Requirements files normally separate the requirements based on where they are needed. This can be useful if different test environments need different requirements.

If that isn't needed, yes then all the development requirements can be in one list somewhere. I'd say it depends on how you test and how you configure CI.

MartinHjelmare (approved):

PR #823, opened 2024-02-07
Add type annotations to examples

emontnemery:


mypy.ini
```diff
@@ -20,6 +20,7 @@ disallow_untyped_defs = true
 warn_return_any = true
 warn_unreachable = true
 exclude = (?x)(
-    ^pychromecast/generated/ # The protobuf autogenerated files are not annotated
+    ^examples/plex_multi_example\.py$
+    | ^pychromecast/generated/ # The protobuf autogenerated files are not annotated
     | ^pychromecast/controllers/plex\.py$
```
emontnemery, line 25: In a separate PR, the plex controller and the example will be type annotated

examples/simple_listener_example.py
```diff
@@ -21,27 +21,27 @@
...
         self.name = name
         self.cast = cast
 
-    def new_media_status(self, status):
+    def new_media_status(self, status: MediaStatus) -> None:
         print("[", time.ctime(), " - ", self.name, "] status media change:")
         print(status)
 
-    def load_media_failed(self, item, error_code):
+    def load_media_failed(self, item: int, error_code: int) -> None:
```
MartinHjelmare, line 44: What does the item represent?

MartinHjelmare (approved):

examples/simple_listener_example.py
emontnemery: It's to identify which item of a queue failed to play, Google calls it `itemId`.

We should update the docstring for `MediaStatusListener` and maybe also rename the parameter to `queue_item` or `queue_item_id` to make it clearer what it is.

PR #832, opened 2024-02-13
Add type annotations to controllers/plex.py

emontnemery:


pychromecast/controllers/plex.py
```diff
@@ -166,8 +181,18 @@ def media_to_chromecast_command(
...
+
+    @property
+    def episode_title(self) -> str | None:
+        """Return episode title."""
+        return self.media_metadata.get("subtitle")
+
+
+# The episode_title property is added to MediaStatus objects
+@property  # type: ignore[misc]
+def episode_title(self: PlexMediaStatus) -> str | None:
```
MartinHjelmare, line 195: This is weird. I'd make this a regular function if it's needed to be separate from the class.

MartinHjelmare (approved):

pychromecast/controllers/plex.py
emontnemery: Yes, it's very weird, I didn't want to clean it up in this PR though

PR #854, opened 2024-02-15
Run rstcheck in CI

emontnemery:


.github/workflows/test.yml
```diff
@@ -35,3 +35,6 @@ jobs:
       - name: Lint with pylint
         run: |
           pylint examples pychromecast
+      - name: Lint with rstcheck
+        run: |
+          rstcheck README.rst
```
MartinHjelmare, line 40: Maybe rewrite it in markdown instead? Almost all our readmes use markdown.

MartinHjelmare (approved):

.github/workflows/test.yml
emontnemery: Yeah, I think that makes sense, we can do that in a follow-up.

PR #861, opened 2024-02-24
Implement dual-stack support

DerEnderKeks:
This PR adds support for IPv6 and thus dual-stack. It contains two changes:

1. Unsetting the `Host` header when requesting the `/setup/eureka_info` endpoint, as I noticed that my NVIDIA Shield blocks any request (empty body with status code 403) when a domain is set in the header, at least when requesting via IPv6.
2. Using a dual-stack socket and IPv6-mapped addresses for the protobuf socket. This remains fully compatible with legacy IPv4, as long as the host can open IPv6 sockets, which any OS from at least the last decade should be able to do.

Let me know if I overlooked anything, but the Youtube example now works fine with both a domain that resolves to only an IPv6 address and a plain IPv4 address.


pychromecast/dial.py
```diff
@@ -90,7 +90,7 @@ def _get_status(
             _LOGGER.debug("Resolved service %s to %s", service, host)
             break
 
-    headers = {"content-type": "application/json"}
+    headers = {"host": "", "content-type": "application/json"}
```
emontnemery, line 93: The purpose of setting host to "" needs to be explained in a comment.
It's mentioned in the PR description that:
> I noticed that my NVIDIA Shield blocks any request (empty body with status code 403) when a domain is set in the header, at least when requesting via IPv6

Does that mean the request is *not* blocked if requesting via IPv4? What is the host header set to in case of requesting the IPv4 and IPv6?

emontnemery:
Please fix the CI issues

pychromecast/dial.py
DerEnderKeks: Added a comment.

I've now also tested this with IPv4 and a domain, and it's blocked too, so it's not IPv6 specific. In addition to the Shield, I tested it with a Google Home Mini, which shows the same behaviour.

The default behaviour with a plain IP is also to send it as the Host header, just like with a domain. Technically, not sending a Host header is [against the spec](https://www.rfc-editor.org/rfc/rfc7230#section-5.4), but so is accepting a request that doesn't have one, which Chromecast devices currently do. To be within spec, any domain would have to be resolved first, and the request would have to be made against the raw IP.

emontnemery: I see. Then please move this change to a separate PR since it's not related to IPv6 support.

emontnemery (changes requested):
Some hosts will have ipv6 disabled, and creating an ipv6 socket will fail, the IPV6 specifics needs to be wrapped with an OSError check with fall back to IPv4.

DerEnderKeks:
The socket will now fall back to IPv4, for systems that are misconfigured to have IPv6 disabled ;)

pychromecast/socket_client.py
```diff
@@ -1133,7 +1142,12 @@ def new_socket() -> socket.socket:
     Try to set SO_REUSEPORT for BSD-flavored systems if it's an option.
     Catches errors if not.
     """
-    new_sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    try:
+        new_sock = socket.socket(socket.AF_INET6, socket.SOCK_STREAM)
+        new_sock.setsockopt(socket.IPPROTO_IPV6, socket.IPV6_V6ONLY, 0)
```
bdraco, line 1147: It would be nice to add a comment here that we want dual stack (thus the `0`)

pychromecast/socket_client.py
```diff
@@ -1133,7 +1142,12 @@ def new_socket() -> socket.socket:
...
+    except OSError:
+        new_sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
```
bdraco, line 1149: It would be nice to add a comment here that some systems will have IPv6 disabled in the kernel (or otherwise broken) so we need to fallback to AF_INET

emontnemery:
@DerEnderKeks please don't amend your previous commit when making changes, it makes it hard to review
(this project always squashes before merge, so no matter how many commits are on your branch, it will end up as a single commit)

DerEnderKeks:
Alright (although I always consider it an unnessacary loss of information when you squash PRs into a single commit).
I fixed the formatting, the check should pass now.

bdraco (approved):

elupus:
Is there any need/benefit of the new happy eyeball stuff you working on @bdraco

bdraco:
All of that work is for asyncio so it wouldn’t apply here

pychromecast/socket_client.py
```diff
@@ -1133,7 +1142,14 @@ def new_socket() -> socket.socket:
...
-    new_sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    try:
+        new_sock = socket.socket(socket.AF_INET6, socket.SOCK_STREAM)
+        # ensuring dual-stack
+        new_sock.setsockopt(socket.IPPROTO_IPV6, socket.IPV6_V6ONLY, 0)
+    except OSError:
+        # falling back to IPv4 on systems without IPv6
+        new_sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+
     new_sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
```
elupus, line 1153: This looks strange to me. We are not making sure the socket type matches the data returned from getaddrinfo. The creation of the socket should be postponed until we have resolved the name using getaddrinfo

elupus (changes requested):

pychromecast/socket_client.py
DerEnderKeks: There isn't really a need to match the socket to the resolved address type, as the dual-stack socket can connect to both types. If it falls back to a v4-only socket and only a v6 address is returned, it would fail to connect either way,

elupus: I dont agree, if we are using getaddrinfo to get the host info, we should respect what it return.

elupus:
Since this sort of died out. I rebased the pull and added iteration over getaddrinfo results. @DerEnderKeks could you test if these changes work for your use case?

PR #864, opened 2024-02-27
Fix hostname support

DerEnderKeks:
> This PR is split from #861

A device will otherwise block the request, when requesting with a hostname instead of the plain IP.

### DIFF
--- a/pychromecast/dial.py
+++ b/pychromecast/dial.py
@@ -90,7 +90,8 @@ def _get_status(
             _LOGGER.debug("Resolved service %s to %s", service, host)
             break
 
-    headers = {"content-type": "application/json"}
+    # unsetting the host header, as requests with a domain would be blocked otherwise
+    headers = {"host": "", "content-type": "application/json"}
 
     if secure:
         url = FORMAT_BASE_URL_HTTPS.format(host) + path


pychromecast/dial.py
```diff
@@ -90,7 +90,8 @@ def _get_status(
             _LOGGER.debug("Resolved service %s to %s", service, host)
             break
 
-    headers = {"content-type": "application/json"}
+    # unsetting the host header, as requests with a domain would be blocked otherwise
```
emontnemery, line 93: Please better explain what this is intended to fix, I can't reproduce any case where this is blocked myself.

Maybe paste some examples using `curl` to craft the working and non-working requests?
Are you sure the problem you see is not caused by an issue in your network configuration?

DerEnderKeks: When requesting the `/setup/eureka_info` endpoint using a hostname (instead of a plain IP), the request is blocked. I can reproduce this both with an Nvidia Shield and a Google Home Mini. I doubt my network influences this, as the devices are on the same L2 and L3 network and can communicate directly. Here are some simple curl commands for reproduction:

```
$ curl https://x.x.x.x:8443/setup/eureka_info -k -v
*   Trying x.x.x.x:8443...
* Connected to x.x.x.x (x.x.x.x) port 8443
* schannel: disabled automatic use of client certificate
* schannel: using IP address, SNI is not supported by OS.
* ALPN: curl offers http/1.1
* ALPN: server did not agree on a protocol. Uses default.
* using HTTP/1.x
> GET /setup/eureka_info HTTP/1.1
> Host: x.x.x.x:8443
> User-Agent: curl/8.4.0
> Accept: */*
>
< HTTP/1.1 200 OK
< Access-Control-Allow-Headers:Content-Type
< Cache-Control:no-cache
< Content-Length:1274
< Content-Type:application/json
<
{...json...}* Connection #0 to host x.x.x.x left intact
```

```
$ curl -H 'Host:' https://homeminitest:8443/setup/eureka_info -k -v
*   Trying x.x.x.x:8443...
* Connected to homeminitest (x.x.x.x) port 8443
* schannel: disabled automatic use of client certificate
* ALPN: curl offers http/1.1
* ALPN: server did not agree on a protocol. Uses default.
* using HTTP/1.x
> GET /setup/eureka_info HTTP/1.1
> User-Agent: curl/8.4.0
> Accept: */*
>
< HTTP/1.1 200 OK
< Access-Control-Allow-Headers:Content-Type
< Cache-Control:no-cache
< Content-Length:1274
< Content-Type:application/json
<
{...json...}* Connection #0 to host homeminitest left intact
```

```
$ curl https://homeminitest:8443/setup/eureka_info -k -v
*   Trying x.x.x.x:8443...
* Connected to homeminitest (x.x.x.x) port 8443
* schannel: disabled automatic use of client certificate
* ALPN: curl offers http/1.1
* ALPN: server did not agree on a protocol. Uses default.
* using HTTP/1.x
> GET /setup/eureka_info HTTP/1.1
> Host: homeminitest:8443
> User-Agent: curl/8.4.0
> Accept: */*
>
< HTTP/1.1 403 Forbidden
< Content-Length:0
<
* Connection #0 to host homeminitest left intact
```

elupus:
I think this might make sense. It is doing ssl authentication. So you setting a host header might trigger the device to try to use some different https certificate.

These device are not using publically signed SSL certs, so that likely causes weird behaviour.

pychromecast/dial.py
```diff
@@ -90,7 +90,8 @@ def _get_status(
...
+    headers = {"host": "", "content-type": "application/json"}
```
elupus, line 94: I think this end up setting it to an empty string. I am not sure that is such a good idea.

"A Host header field must be sent in all HTTP/1.1 request messages. A [400 Bad Request](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400) status code may be sent to any HTTP/1.1 request message that lacks or contains more than one Host header field."

So it likely need to be set to the IP address of the request.

elupus (changes requested):
While i think the need might be real, i think the solution need to be adjusted.

pychromecast/dial.py
emontnemery: I would still like an example showing where this is a problem and how this PR fixes it.

elupus: I suspect just setting a manual dns (ie not mdns) hostname for a chromecast instead of an IP would cause it.

elupus: Just realized that it might only be possible with the changes for ipv6 that was posted that does a getaddrinfo lookup.

DerEnderKeks: My usecase is indeed to have a custom DNS name for a cast device, as I can't use mDNS in my setup, since the cast devices are in a different L2/L3 network than Home Assistant.

I do agree with setting the `Host` header to the resolved IP, that's a cleaner way to solve this.

DerEnderKeks:
> I think this might make sense. It is doing ssl authentication. So you setting a host header might trigger the device to try to use some different https certificate.
> 
> These device are not using publically signed SSL certs, so that likely causes weird behaviour.

@elupus This is not related to TLS (my curl examples all ignore invalid certificates through `-k`), but more likely some kind of DNS rebinding protection in the Cast web server, thus the `403` status code.

pychromecast/dial.py
tbnobody: As mentioned here: https://github.com/home-assistant/core/issues/135024#issuecomment-2724875127

Adding the IP of the chromecast in the host header fixes the issue

emontnemery: I changed the PR to do something inbetween; if the HTTP get fails with FORBIDDEN, we retry with an empty `host` header.

The reason I suggest this simplified approach is to avoid doing the name resolution via `socket.getaddrinfo`, and then iterating over the returned list.

pychromecast/dial.py
```diff
@@ -101,9 +100,24 @@ def _get_status(
...
+    try:
+        req = urllib.request.Request(url, headers=headers)
+        with urllib.request.urlopen(req, timeout=timeout, context=context) as response:
+            data = response.read()
+    except urllib.error.HTTPError as err:
+        if err.code != HTTPStatus.FORBIDDEN:
+            raise
+        # We may be blocked because we're connecting to a hostname not matching the hostname
+        # published by the chromecast device via mDNS. Do another attempt with an empty
+        # host header.
```
emontnemery, line 112: Maybe I'm lying here; I'm assuming the device will accept the hostname it itself publishes. Or do devices only accept no host header or host header set to the stringified IP?

emontnemery: I tested on one of my own devices, and it doesn't accept the HTTP request also when `host` is set to the hostname it itself publishes.

pychromecast/dial.py
```diff
@@ -74,6 +77,31 @@ def _get_host_from_zc_service_info(
...
 
+@contextmanager
+def _urlopen(
+    url: str, timeout: float, context: ssl.SSLContext | None
+) -> Generator[Any, None, None]:
+    """Help open an URL."""
+    headers = {"content-type": "application/json"}
+    try:
+        req = urllib.request.Request(url, headers=headers)
+        yield urllib.request.urlopen(req, timeout=timeout, context=context)
```
elupus, line 88:
```suggestion
        with urllib.request.urlopen(req, timeout=timeout, context=context) as response
            yield response
```

elupus (approved):
I would have done as in the suggestion. But this is fine!

pychromecast/dial.py
```diff
@@ -75,6 +77,31 @@ def _get_host_from_zc_service_info(
     return (host, port)
 
 
+@contextmanager
```
elupus, line 80:
```suggestion
```

pychromecast/dial.py
```diff
@@ -75,6 +77,31 @@ def _get_host_from_zc_service_info(
...
+def _urlopen(
+    url: str, timeout: float, context: ssl.SSLContext | None
+) -> Generator[Any, None, None]:
+    """Help open an URL."""
+    headers = {"content-type": "application/json"}
+    try:
+        req = urllib.request.Request(url, headers=headers)
+        yield urllib.request.urlopen(req, timeout=timeout, context=context)
```
elupus, line 88:
```suggestion
        return urllib.request.urlopen(req, timeout=timeout, context=context)
```

pychromecast/dial.py
```diff
@@ -75,6 +77,31 @@ def _get_host_from_zc_service_info(
...
+        # instead of to an IP address resolved by mDNS, cast devices will reject the
+        # request when the address is set to the hostname. Do another attempt with an
+        # empty host header.
+        # Note: This is a simplified approach to not have to deal with name resolution
+        # in pychromecast. If devices reject the empty host header we need to do name
+        # resolution and instead set the host header to the string version of the IP.
+        _LOGGER.debug("Failed to fetch %s, retrying with empty host header", url)
+        headers["host"] = ""
+        req = urllib.request.Request(url, headers=headers)
+        yield urllib.request.urlopen(req, timeout=timeout, context=context)
```
elupus, line 102:
```suggestion
        return urllib.request.urlopen(req, timeout=timeout, context=context)
```

elupus (changes requested):

PR #865, opened 2024-03-01
Migrate to using efficient I/O multiplexing selectors

bdraco:
When looking at a profile of a user with ~25 cast devices, I found that the threads wake up often take enough time away from the event loop that it made the system sluggish.

Migrate to using the most efficient [selector](https://docs.python.org/3/library/selectors.html) available


bdraco:
I don't have a cast device to test this with so ordered one, and I'll have to wait for one to show up + test + adjust before I can mark this ready for review

pychromecast/socket_client.py
```diff
@@ -64,12 +64,11 @@
 CONNECTION_STATUS_FAILED_RESOLVE = "FAILED_RESOLVE"
 # The socket connection was lost and needs to be retried
 CONNECTION_STATUS_LOST = "LOST"
-# Check for select poll method
-SELECT_HAS_POLL = hasattr(select, "poll")
 
 HB_PING_TIME = 10
 HB_PONG_TIME = 10
-POLL_TIME_BLOCKING = 5.0
+# Try not to wake up too often
```
bdraco, line 70: 60 seconds / 5 s interval = 12 thread wakes per minute
25 device * 12 thread wakes per minute = 300 idle thread wakes per minute

pychromecast/socket_client.py
```diff
@@ -64,12 +64,11 @@
...
+POLL_TIME_BLOCKING = 300.0
```
bdraco, line 71: At 300s block time:

~ 5 idle thread wakes per minute with 25 devices

bdraco:
I bought a chromecast device to test with

Verified the following are ok

- add device
- add/remove config entry
- hard power off device, and recover
- restart HA

pychromecast/socket_client.py
emontnemery: I don't think there should be any timeout here at all, we have a socketpair with the purpose of waking up the thread in a timely manner and we should have removed the timeout when we added the socketpair.

pychromecast/socket_client.py
```diff
@@ -215,6 +214,11 @@ def __init__(
         self.connecting = True
         self.first_connection = True
         self.socket: socket.socket | ssl.SSLSocket | None = None
+        self.selector = selectors.DefaultSelector()
+        self.self_selector_key = self.selector.register(
+            self.socketpair[0], selectors.EVENT_READ
+        )
+        self.socket_selector_key: selectors.SelectorKey | None = None
```
emontnemery, line 221: Can we try to come up with more descriptive names? The socketpair is used to wake up the thread, the socket is used to talk to a remote cast device.

bdraco:
> I don't think there should be any timeout here at all, we have a socketpair with the purpose of waking up the thread in a timely manner and we should have removed the timeout when we added the socketpair.
> Edit: PR https://github.com/home-assistant-libs/pychromecast/pull/877 removes the timeouts

Awesome. I was hoping you would say that 👍

pychromecast/socket_client.py
bdraco: `self_selector_key` -> `wakeup_selector_key`
`socket_selector_key` -> `remote_selector_key`

Does that make more sense ?

emontnemery: yeah, that's more clear I think

emontnemery (approved):
LGTM, thanks @bdraco 👍

PR #895, opened 2024-04-17
Expose 'manufacturer' and 'product_name' properties in 'Chromecast' class.

pjgrochowski:
Expose 'manufacturer' and 'product_name' properties in 'Chromecast' class.

**[Justification]:**
Field '**model_name**' evaluates to '**Chromecast**' for all device models.
Field '**product_name**' allows to actually distinguish between Chromecast models:
- Chromecast 3 => 'caprica'
- Chromecast Ultra => 'steak'
- Chromecast Android TV => 'sabrina'

### DIFF
--- a/pychromecast/__init__.py
+++ b/pychromecast/__init__.py
@@ -395,12 +395,28 @@ class Chromecast(CastStatusListener):
         return f"{self.socket_client.host}:{self.socket_client.port}"
 
     @property
+    def manufacturer(self) -> str:
+        """Returns the manufacturer of the Chromecast device."""
+        if TYPE_CHECKING:
+            # get_cast_type is guaranteed to return a CastInfo with a non-None model
+            assert self.cast_info.manufacturer is not None
+        return self.cast_info.manufacturer
+
+    @property
     def model_name(self) -> str:
         """Returns the model name of the Chromecast device."""
         if TYPE_CHECKING:
             # get_cast_type is guaranteed to return a CastInfo with a non-None model
             assert self.cast_info.model_name is not None
         return self.cast_info.model_name
+
+    @property
+    def product_name(self) -> str:
+        """Returns the product name of the Chromecast device."""
+        if TYPE_CHECKING:
+            # get_cast_type is guaranteed to return a CastInfo with a non-None model
+            assert self.cast_info.product_name is not None
+        return self.cast_info.product_name
 
     @property
     def cast_type(self) -> str:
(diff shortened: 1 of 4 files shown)


elupus (dismissed):
I think this make sense. Would map to new field in HA too.

pychromecast/__init__.py
```diff
@@ -394,6 +394,14 @@ def uri(self) -> str:
         """Returns the device URI (ip:port)"""
         return f"{self.socket_client.host}:{self.socket_client.port}"
 
+    @property
+    def manufacturer(self) -> str:
+        """Returns the manufacturer of the Chromecast device."""
+        if TYPE_CHECKING:
+            # get_cast_type is guaranteed to return a CastInfo with a non-None model
```
emontnemery, line 401: The comment is misleading

pychromecast/__init__.py
```diff
@@ -402,6 +410,14 @@ def model_name(self) -> str:
             assert self.cast_info.model_name is not None
         return self.cast_info.model_name
 
+    @property
+    def product_name(self) -> str:
+        """Returns the product name of the Chromecast device."""
+        if TYPE_CHECKING:
+            # get_cast_type is guaranteed to return a CastInfo with a non-None model
```
emontnemery, line 417: Same

emontnemery (commented):
I'm not fully sure about the benefit of this, why is it better to expose this in a property, instead of accessing via the public `cast_info`?

elupus (commented):
Hmm. Didn't realize cast_info was public already. Seem pointless yes. One can wonder why we expose the others then.

elupus (changes requested):
Missing addition of the new field here:
https://github.com/home-assistant-libs/pychromecast/pull/895/files#diff-c7c448ef8026704bcb4f958857c133335170bbeda706cea0bfeaa033ad829a7dR265
and 
https://github.com/home-assistant-libs/pychromecast/pull/895/files#diff-c7c448ef8026704bcb4f958857c133335170bbeda706cea0bfeaa033ad829a7dR251
and
dial.py

emontnemery:
> Hmm. Didn't realize cast_info was public already. Seem pointless yes. One can wonder why we expose the others then.

It's due to some old design choice to forward selected attributes and methods from sub-objects to the `Pychromecast` object.

See for example here:
https://github.com/home-assistant-libs/pychromecast/blob/491789843f2bde6c21fe428c387d4064c425a191/pychromecast/__init__.py#L344-L356

It's currently applied very inconsistently; new functionality has not been forwarded like this for several years.

Notice how `media_controller.play_media` is forwarded, but no other methods of it are.

elupus:
Since no feedback from developer and this change is incomplete, let's close this for now.

PR #946, opened 2024-09-04
Fix errors found by pylint 3.2.7

emontnemery:


pychromecast/socket_client.py
```diff
@@ -633,6 +633,10 @@ def _run_once(self) -> int:
             return 0
 
         # See if any handlers will accept this message
+
+        # Pylint says data is possibly unassigned. That's not the case though,
+        # because if it is, message is None and we've returned just above
+        # pylint: disable-next=possibly-used-before-assignment
         self._route_message(message, data)
```
emontnemery, line 640: Another pair on eyes on this would be great in case I miss something

elupus (approved):
Ser bättre ut

PR #987, opened 2024-11-13
Avoid logging graceful disconnect as error

elupus:
The chromecasts restart each night, and seem to
generally do this by doing a socket shutdown of
write leading to a EOF received by clients.

No need to log this as an error

### DIFF
--- a/pychromecast/error.py
+++ b/pychromecast/error.py
@@ -11,6 +11,10 @@ class ChromecastConnectionError(PyChromecastError):
     """When a connection error occurs within PyChromecast."""
 
 
+class ChromecastConnectionClosed(PyChromecastError):
+    """When a connection was closed by remote device."""
+
+
 class PyChromecastStopped(PyChromecastError):
     """Raised when a command is invoked while the Chromecast's socket_client
     is stopped.
--- a/pychromecast/socket_client.py
+++ b/pychromecast/socket_client.py
@@ -33,6 +33,7 @@ from .controllers.receiver import CastStatus, CastStatusListener, ReceiverContro
 from .dial import get_host_from_service
 from .error import (
     ChromecastConnectionError,
+    ChromecastConnectionClosed,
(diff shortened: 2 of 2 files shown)


elupus:
Hmm. I wonder should we set stop()? Will something restart the thread again when a service is spotted?

emontnemery:
> Not verified this, but if they behave correctly (which I suspect they do) they remove their mdns entry either before or more likely after they disconnect.

I've had lots of problems with this before. We need a solution which works well both when the restart is unexpected, for example a power cut, and when the device reboots normally.

emontnemery:
> Hmm. I wonder should we set stop()? Will something restart the thread again when a service is spotted?

Please elaborate. Which thread are you talking about?

emontnemery:
I think debug level is way too quiet in case of an unexpected disconnect, info would be better.

elupus:
This is not unexpected. Its a gracefull requested shutdown from remote, which occurs nightly. We can have it at INFO, since that is hidden in HA by default.

elupus:
The point is, when we get a graceful disconnect (eof from remote), we should delay the reconnect untill either some set timeout or a mdns broadcast is seen.

In case of a non graceful disconnect (socket RST), we can try to reconnect directly.

elupus:
> > Hmm. I wonder should we set stop()? Will something restart the thread again when a service is spotted?
> 
> Please elaborate. Which thread are you talking about?

If i understand correctly each client handling a socket is a separate thread. To wait for next mdns event i was thinking of stopping down the current client thread and restarting on next announce. But I'm not fully following the code yet. So baby steps.

gvdhoven:
The issue (reconnect every night) actually causes Home Assistant to crash. Also, a 5 second delay before reconnecting might be a bit to soon.

gvdhoven:
See here: https://github.com/home-assistant/core/issues/135428

PR #990, opened 2024-11-20
Set ciphers to DEFAULT in socket_client.py to connect with a wider range of chromecast devices

eljefe80:
Add context.set_ciphers('DEFAULT') to allow older ciphers to work


emontnemery:
Please motivate why this change is needed

eljefe80:
HI @emontnemery!

As described in https://github.com/home-assistant-libs/pychromecast/issues/963, there is at least one closed source chromecast client that is using older cyphers that fail connection attempts as is. I wouldn't say it's needed, but personally, it would be nice to not have to edit this file every time a new home assistant container is deployed.

Thanks,
Jeff

PR #997, opened 2024-12-12
Advertise typing support

C0rn3j:
This project seems great with handling typing, but it does not announce its support in a conformant fashion, so tools like Pyright do not pick up on it.

Add `Typing` to pyproject.toml and create `py.typed` in the module directory.

https://typing.readthedocs.io/en/latest/spec/distributing.html#packaging-typed-libraries


emontnemery (approved):
Thanks a lot @C0rn3j 👍

PR #1023, opened 2025-02-15
Fix typos in Plex example

vesper8:
I got confused because I typed "simple" in the demo arg below after reading the doc block at the top and it wouldn't work.. I then realized that the code below is looking for "single" not "simple"


MartinHjelmare (approved):
Thanks!

PR #1048, opened 2025-04-04
Rewritten library using asyncio

albaintor:
Hi,

I rewrote the library using asyncio instead of threads. I had issues with the old behaviour and it was not suitable to use with an async program.

I tested it successfully on my side. Of course there are some breaking changes :
Instead of calling `wait` method, we use the connect method directly like this :
```python
chromecast=pychromecast.get_chromecast_from_host((host, 8009, None, None, None))
await chromecast.connect(timeout=5)
```

As you can see the code is much more simpler as a lot of work is performed directly in standard wrappers of asyncio.

### DIFF
--- a/.github/workflows/pythonpublish.yml
+++ b/.github/workflows/pythonpublish.yml
@@ -13,7 +13,7 @@ jobs:
     steps:
     - uses: actions/checkout@v4.2.2
     - name: Set up Python
-      uses: actions/setup-python@v5.4.0
+      uses: actions/setup-python@v5.5.0
       with:
         python-version: '3.11'
     - name: Install dependencies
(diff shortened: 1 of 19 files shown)


elupus:
Sorry pressed wrong button. I like the effort, but i wonder how we can handle this in a non breaking way for current users of the library.

albaintor:
There is indeed some breaking changes but not so much, only those 2  : 

- the initialization of the connection should be replaced with an await prefix in an async method
- all the calls to commands should be prefixed with await in an async method

The rest remains unchanged, including all the registrations to the listeners.

This is a great benefit I think especially for HA which works only with a single event loop in a single thread.
I experienced random deadlocks with the current implementation

PR #1051, opened 2025-04-10
Fix status parsing of isActiveInput and isStandby

incase:
These were broken by commit 8b16e86c7ec4b9ab4e86bfd632864d32b82da111, which introduced the new status_data dict, replacing a shadowing "data" dict, but missing these two references.

(https://github.com/home-assistant-libs/pychromecast/commit/8b16e86c7ec4b9ab4e86bfd632864d32b82da111)


elupus (approved):
Ah this explain some of the oddness ive seen.

PR #1162, opened 2026-02-09
Allow connecting to chromecast over IPv6

Mathis-6:
Currently, we can only connect to the chromecasts over IPv4 because `socket.AF_INET` is hardcoded for the socket.
This PR adds an automatic detection to choose from AF_INET6 or AF_INET, depending if the address contains colons or not

### DIFF
--- a/pychromecast/socket_client.py
+++ b/pychromecast/socket_client.py
@@ -173,6 +173,7 @@ class SocketClient(threading.Thread, CastStatusListener):
         retry_wait: float | None,
         services: set[HostServiceInfo | MDNSServiceInfo],
         zconf: zeroconf.Zeroconf | None,
+        ip_family: socket.AddressFamily = socket.AF_INET,
     ) -> None:
         super().__init__()
 
@@ -192,6 +193,7 @@ class SocketClient(threading.Thread, CastStatusListener):
 
         self.host = "unknown"
         self.port = 8009
+        self.ip_family = ip_family
 
         self.source_id = "sender-0"
         self.stop = threading.Event()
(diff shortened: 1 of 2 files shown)


pychromecast/socket_client.py
```diff
@@ -1085,15 +1086,14 @@ def receive_message(self, message: CastMessage, data: dict) -> bool:
         return False
 
 
-def new_socket() -> socket.socket:
+def configure_socket(_socket: socket.socket) -> None:
```
mitchellrj: Nit: by [PEP 8 convention](https://pep8.org/#descriptive-naming-styles) this should be a trailing underscore, rather than a leading one (indicating internal use).

mitchellrj:
The connection fails for me due to the input being `[abcd:1234::5789]` [host format](https://datatracker.ietf.org/doc/html/rfc2732) rather than `abcd:1234::5789` IP format.

This might not be a problem for everyone but it certainly will be for some people. If this is working for you, are you using `glibc`? What version?

#1199 had a workaround for this, but better to apply that fix here than bring in the rest of the unnecessary complexity I added.

Mathis-6:
@mitchellrj ipv6 addresses enclosed in square brackets don't work for me too.
Why do you need to pass an ipv6 address enclosed in brackets ? Can't you remove them before passing the address to pychromecast ?

mitchellrj:
Zeroconf / mdns discovery returns the address in hostname format, which is where this breaks down for me.

Mathis-6:
Based on my fork could you write a simple PR to address this issue ? I don't think you need 200 LoC now that most of the IPv6 part is working.

PR #1164, opened 2026-02-21
Prevent tight loop if Plex app failed to launch

mdeneen:
I've been tracking this one down for a while in Home Assistant.

Occasionally, when playing Plex media on a Google Home Mini, Home Assistant would get into a very tight loop (see https://github.com/home-assistant/core/issues/161641) if there was a failure launching the Plex app on the device.  This patch fixes that behavior by logging that it has happened, setting the event, and returning instead of raising an exception.

Prior to this patch, once this exception happened, Home Assistant would get into a tight loop and continuously log errors until all disk space was consumed.

I'm not super familiar with how pychromecast works, but understood enough to make it not fail.  :-)

Please review and apply if appropriate.

### DIFF
--- a/pychromecast/controllers/plex.py
+++ b/pychromecast/controllers/plex.py
@@ -488,7 +488,9 @@ class PlexController(BaseController):
 
         def app_launched_callback(msg_sent: bool, _response: dict | None) -> None:
             if not msg_sent:
-                raise RequestFailed("PlexController.play_media")
+                self.logger.warning("PlexController.play_media: unable to launch app, device unavailable?")
+                self.play_media_event.set()
+                return
             try:
                 self._send_start_play(media, **kwargs)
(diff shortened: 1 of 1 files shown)


elupus (changes requested):
This does not seem right. Hone assistant should not catch that exception and retry indefinitely.

mdeneen:
I can assure you that it does.

mdeneen:
Second attempt has been made in #1208

elupus:
So i think we should swallow the exception for the callbacks, but it should be done when calling the callback not inside the callback.

Since the callbacks are run from the connection task, but callbacks are registered from requests i think, they have no business taking down the connection task. So wherever we call the callbacks, we should only log exceptions, not let the propagate.

PR #1199, opened 2026-04-25
feat: IPv6-only support

mitchellrj:
For devices which only have IPv6 addresses, this change allows creation of IPv6 sockets or IPv4 sockets, depending on what is available.

It does not allow for Happy Eyeballs implementation on dual-stack, which would require a much larger refactor.

`[abcd:1234::6789]` notation is retained for HTTP connections (required to distinguish the address from the port), but stripped out for the TCP socket connection, since some implementations of `getaddrinfo` do not support this notation as input, only the raw IPv6 address.

### DIFF
--- a/pychromecast/const.py
+++ b/pychromecast/const.py
@@ -2,6 +2,11 @@
 Chromecast constants
 """
 
+from typing import TypeAlias
+
+# Type definitions / aliases
+HostnameType: TypeAlias = str
+
 # Regular chromecast, supports video/audio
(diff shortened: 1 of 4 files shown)


elupus:
Did you see: https://github.com/home-assistant-libs/pychromecast/pull/1162 (ive not had time to look at it)

mitchellrj:
Well that PR is much neater than mine! I missed that in my search of issues.

elupus:
Could you test it and post a review?

PR #1242, opened 2026-07-12
Swallow and log unhandled exceptions when callbacks are called

mdeneen:
This is my third attempt at this, with prior attempts being #1164 and #1208, discussed in https://github.com/home-assistant/core/issues/161641

This change catches the exception at the socket client level and should function with any handler that happens to throw an exception which is unhandled in a callback.

I've watched my logs and can confirm that the exception is thrown from the Plex handler as it tries to issue the callbacks on reconnect, caught and logged in the socket client, and the connection is successfully re-established.


elupus (approved):
Looks good thanks!

